PR #29, opened 2018-01-30
rake indexer for ES

kjgarza:
closes datacite/levriero#14


kjgarza:
So a custom API method for bulk indexing? ok

mfenner:
It can be a rake task. But I would use the rake task just to trigger, and have all the logic in the model. API method can come later.

app/models/provider.rb
```diff
@@ -119,6 +122,16 @@ def user_url
     ENV["VOLPINO_URL"] + "/users?provider-id=" + symbol.downcase
   end
 
+  def self.push_to_index
+    data = self.all
```
mfenner: I would avoid `self.all` together with `.each`. You would load everything into memory and this does not scale well. Better to use `find_each`: https://apidock.com/rails/ActiveRecord/Batches/find_each

app/models/client.rb
```diff
@@ -123,6 +123,17 @@ def user_url
     ENV["VOLPINO_URL"] + "/users?client-id=" + symbol.downcase
   end
 
+  def self.push_to_index
+    self.find_each do |client|      
+      params = { "data" => { "type" => "clients", "attributes" => client.attributes } }
+      params["data"]["attributes"]["contact-email"]= params["data"]["attributes"]["contact_email"]
```
mfenner: This can be done simpler, along the lines of https://github.com/datacite/volpino/blob/master/app/controllers/application_controller.rb#L18-L20.

app/jobs/elasticsearch_job.rb
```diff
@@ -0,0 +1,29 @@
+class ElasticsearchJob < ActiveJob::Base
+    queue_as :default
+
+    rescue_from ActiveJob::DeserializationError, Faraday::TimeoutError do
+        retry_job wait: 5.minutes, queue: :default
+    end
+
+    def perform(data, operation)
+      options = { content_type: 'application/vnd.api+json', accept: 'application/vnd.api+json', bearer: ENV['JWT_TOKEN'] }
```
mfenner: I think this is the only place where we use a JWT token. How do we generate the token?

app/models/provider.rb
```diff
@@ -119,6 +122,15 @@ def user_url
     ENV["VOLPINO_URL"] + "/users?provider-id=" + symbol.downcase
   end
 
+  def self.push_to_index
+    self.find_each do |provider|    
+      params = { "data" => { "type" => "providers", "attributes" => provider.attributes } }
+      params["data"]["attributes"]["updated"]= params["data"]["attributes"]["updated"].to_s
```
mfenner: see above.

config/application.rb
```diff
@@ -46,6 +46,7 @@
 ENV['MYSQL_HOST'] ||= "mysql"
 ENV['MYSQL_PORT'] ||= "3306"
 ENV['TRUSTED_IP'] ||= "10.0.40.1"
+ENV['LEVRIERO_URL'] ||= "https://api.datacite.org"
```
mfenner: Is this going to work?

mfenner:
I would add at least basic test coverage.

config/application.rb
kjgarza: Well, actually here is where I get to the POST method discussion in https://api.datacite.org. If I make a POST request in a consolidated API, then it will write to the database, all the time. unless we use admin domain for ES:POST, right?

PR #35, opened 2018-02-12
added prefixes for filtering

kjgarza:
I thought there might be 3 ways to add prefixes relationship at this moment.

1. Include them in the rake task
2. Make calls to app.datacite.org for prefixes relationships
3. implement the prefixes model

I leaned for (1) as we can move to the number (3) later with little effort involved.


kjgarza:
this was done in another pullrequest already

PR #47, opened 2018-02-28
register DOI in the handle system

mfenner:
This pull request adds DOI registration in the handle system. It does so by calling the MDS API. This is the only MDS API functionality used by the Lupo API, and can over time be replaced by a direct call to the Handle REST API.

One requirement is basic authentication, and only clients (not providers and admins) have permissions. This is implemented by storing the password in the jwt for clients. Not ideal, but probably the best workaround for enabling DOI Fabrica integration.

### DIFF
--- a/app/models/concerns/helpable.rb
+++ b/app/models/concerns/helpable.rb
@@ -7,6 +7,27 @@ module Helpable
   included do
     include Bolognese::DoiUtils
     include Cirneco::Utils
+    include Cirneco::Api
+
+    attr_accessor :username, :password
+
+    def register_url
+      return OpenStruct.new(body: { "errors" => [{ "title" => "Username or password missing" }] }) unless username.present? && password.present?
+
+      response = put_doi(doi, url: url,
+                         username: username,
+                         password: password,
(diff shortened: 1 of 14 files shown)


app/models/concerns/helpable.rb
```diff
@@ -7,6 +7,27 @@ module Helpable
   included do
     include Bolognese::DoiUtils
     include Cirneco::Utils
+    include Cirneco::Api
+
+    attr_accessor :username, :password
+
+    def register_url
```
kjgarza: So its uses cirneco and in place uses MDS, right? @mfenner

mfenner: Yes, cirneco to update the URL in the handle system via MDS API.

mfenner: I haven't tested the full integration, but, together with using basic auth in `bracco`, it solves the issue of talking to the handle server. It also means we can remove a similar call in `cheetoh`, so that `cheetoh` is no longer talking to the MDS API, but only to `lupo`.

PR #55, opened 2018-03-29
Bug xml differs from input

kjgarza:
unfortunately this change break test for /validate
I put it such a way that it will never ask bolognese to create the XML. It should always ask for the database metadata.

However, that breaks the validate action. This is a new action, i am unsure what do we use it for. I will need to check further but I will manage the issue from here. 

this is related to datacite/bolognese#23

### DIFF
--- a/app/models/concerns/crosscitable.rb
+++ b/app/models/concerns/crosscitable.rb
@@ -74,14 +74,14 @@ module Crosscitable
     end
 
     def xml
-      datacite
+      current_metadata.xml
     end
 
     def xml=(value)
       return nil unless value.present?
 
       input = Base64.decode64(value).force_encoding("UTF-8")
-
+      
       options = {
         doi: doi,
         sandbox: !Rails.env.production?,
@@ -103,7 +103,7 @@ module Crosscitable
 
       @schema_version = bolognese.schema_version || "http://datacite.org/schema/kernel-4"
       @from = bolognese.from
-      @raw = bolognese.raw
+      @raw = value
 
       metadata.build(doi: self, xml: datacite, namespace: @schema_version)
     rescue NoMethodError, ArgumentError => exception
--- a/spec/requests/dois_spec.rb
+++ b/spec/requests/dois_spec.rb
@@ -280,6 +280,43 @@ describe "dois", type: :request do
       end
     end
 
+    context 'when the request contains subject scheme is valid' do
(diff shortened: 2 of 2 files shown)


app/models/concerns/crosscitable.rb
```diff
@@ -74,14 +74,14 @@ def datacite
     end
 
     def xml
-      datacite
+      current_metadata.xml
```
mfenner: This changes breaks the conversion of other input formats to DataCite XML.

app/models/concerns/crosscitable.rb
```diff
@@ -103,7 +103,7 @@ def xml=(value)
 
       @schema_version = bolognese.schema_version || "http://datacite.org/schema/kernel-4"
       @from = bolognese.from
-      @raw = bolognese.raw
+      @raw = value
```
mfenner: This change doesn't work if `value` is metadata in another format, and not DataCite XML.

spec/requests/dois_spec.rb
```diff
@@ -280,6 +280,43 @@
...
+      end
+
+      before { post '/dois', params: valid_attributes.to_json, headers: headers }
+
+      it 'creates a Doi' do
+        puts json.inspect
+        expect(json.dig('data', 'attributes', 'url')).to eq("http://www.bl.uk/pdf/patspec.pdf")
+        expect(json.dig('data', 'attributes', 'doi')).to eq("10.4122/10703")
+        expect(json.dig('data', 'attributes', 'title')).to eq("Submitted chemical data for InChIKey=YAPQBXQYLJRXSA-UHFFFAOYSA-N")
+        expect(json.dig('data', 'attributes', 'xml')).to eq(xml)
```
kjgarza: @mfenner Does this spec reflect correct behaviour tho? Meaning, On creating a DOI resource we ought to expect that XML field is always the same?

mfenner: Almost. The DOI can be different, e.g. in cases of randomly generated DOIs.

app/models/concerns/crosscitable.rb
kjgarza: ok; So we are clear on that the defualt storage format is always DataCite XML. Good. Then, bolognese.raw is not working when the original format is datacite-xml

app/models/concerns/crosscitable.rb
kjgarza: I'm unclear on this one. If the default storage format is DataCite XML, why do we need a conversion step in the getter of the XML field ??

app/models/concerns/crosscitable.rb
mfenner: > bolognese.raw is not working when the original format is datacite-xml

Why is that?

app/models/concerns/crosscitable.rb
mfenner: You are right. I was thinking that in some cases we might do a conversion of the DataCite XML, but not here.

app/models/concerns/crosscitable.rb
kjgarza: I was thinking that this might be true because the spec (file below) was always failing in the comparison of XML. however, upon further checking, it seems that Bolognese (something in LUPO) `prettify` XML (adding tabs) and the difference on XML check comes from there.  Therefore I think comparing XML on input vs XML on output is not correct (i'll check my spec) . I still think the getter (file above) is the main issue: there shouldn't be any conversion. I'll focus on that.

mfenner:
@kjgarza I made pull request #57 as an alternative. The reason I didn't update this pull request is that there are already a number of other changes (mostly updating to Rails 5.2RC2).

PR #57, opened 2018-04-01
Keep datacite xml

mfenner:
Better handling of DataCite XML, so that it is not overwritten by newly generated xml.

### DIFF
--- a/app/serializers/doi_serializer.rb
+++ b/app/serializers/doi_serializer.rb
@@ -2,7 +2,7 @@ class DoiSerializer < ActiveModel::Serializer
   include Bolognese::Utils
   include Bolognese::DoiUtils
 
-  attributes :doi, :prefix, :suffix, :identifier, :url, :creator, :title, :publisher, :resource_type_subtype, :description, :version, :metadata_version, :schema_version, :state, :is_active, :reason, :landing_page, :published, :registered, :updated
+  attributes :doi, :prefix, :suffix, :identifier, :url, :creator, :title, :publisher, :resource_type_subtype, :description, :version, :metadata_version, :schema_version, :state, :is_active, :reason, :landing_page, :xml, :published, :registered, :updated
 
   belongs_to :client, serializer: ClientSerializer
   belongs_to :provider, serializer: ProviderSerializer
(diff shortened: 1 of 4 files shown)


app/serializers/doi_serializer.rb
```diff
@@ -2,7 +2,7 @@ class DoiSerializer < ActiveModel::Serializer
   include Bolognese::Utils
   include Bolognese::DoiUtils
 
-  attributes :doi, :prefix, :suffix, :identifier, :url, :creator, :title, :publisher, :resource_type_subtype, :description, :version, :metadata_version, :schema_version, :state, :is_active, :reason, :landing_page, :xml, :published, :registered, :updated
+  attributes :doi, :prefix, :suffix, :identifier, :url, :creator, :title, :publisher, :resource_type_subtype, :description, :version, :metadata_version, :schema_version, :state, :is_active, :reason, :landing_page, :published, :registered, :updated
```
kjgarza: But don't we need this in bolognesa and content-negotiation ?https://github.com/datacite/bolognese/blob/62e94949d4363cb55bfe9cf4e97715b940866422/lib/bolognese/readers/datacite_reader.rb#L18

mfenner: I added it back in for compatibility. But I want to rewrite how bolognese and content-negotiation are fetching metadata - content negotiation is too slow, and the main reason is always this metadata conversion overhead.

kjgarza: I actually think the bug I was fixing is partially related to this part. Because LUPO seem to regenerate XML using Bolognese and put it in this field in the serializer and Bolognese calls LUPO to get the XML from the same field. In some cases its like a circular relationship.

mfenner: Content negotiation would be faster if we either fetch crosscite.json or datacite.xml from `lupo`, depending on the request.

mfenner: `bracco` isn't using the `xml` attribute in the `lupo` API response anymore, but instead uses content negotiation for DataCite XML.

mfenner: @kjgarza yes, it is convoluted and circular, as `bolognese` fetches DataCite XML from `lupo` which uses `bolognese`...

mfenner:
Closes #56. More work is needed.

PR #109, opened 2018-09-04
Adding new column for storing link check result

richardhallett:
This is a migration to add a new column so we can store a result details from our link checker.


richardhallett:
@kjgarza Can you expand upon what you mean? I don't think we're using the version of MYSQL that supports a specific JSON type? is this a special thing in Rails for specifying JSON?

kjgarza:
@richardhallett when we talked I thought you mentioned it was going to be a json. I thought 5.7 supported this https://github.com/datacite/mastino/blob/43482cc7079bd98b406682546018b0f8a6bd838a/prod-eu-west/data-storage/mysql/main.tf#L3

PR #114, opened 2018-10-05
Add support for storing link check result

richardhallett:
This is to allow requests to be made to store raw results found by our crawler to avoid losing potential information or for later processing.
Note I've not indexed it in ES as it's not intended to be searched upon.

### DIFF
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -36,7 +36,7 @@ class DoisController < ApplicationController
       end
 
       page = params[:page] || {}
-      if page[:size].present? 
+      if page[:size].present?
         page[:size] = [page[:size].to_i, 1000].min
         max_number = page[:size] > 0 ? 10000/page[:size] : 1
       else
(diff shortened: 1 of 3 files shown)


app/controllers/dois_controller.rb
```diff
@@ -384,7 +384,7 @@ def set_include
...
+    attributes = [:doi, "confirm-doi", :identifier, :url, :title, :publisher, :published, :created, :prefix, :suffix, "resource-type-subtype", "last-landing-page", "last-landing-page-status", "last-landing-page-status-check", "last-landing-page-status-result", "last-landing-page-content-type", "content-url", "content-size", "content-format", :description, :license, :xml, :validate, :source, :version, "metadata-version", "schema-version", :state, "is-active", :reason, :registered, :updated, :mode, :event, :regenerate, :client, "resource_type", author: [:type, :id, :name, "given-name", "family-name", "givenName", "familyName"]]
```
mfenner: `last-landing-page-status-result` is not atomic, but a hash. Should be handled different (see line 388 for examples).

mfenner:
@richardhallett LGTM. Do you want to merge?

PR #121, opened 2018-10-19
Update remove miliseconds from created and updated

kjgarza:
more detail in datacite/search#170


kjgarza:
forgot to run it locally, wait a second

mfenner:
Right, you want to update the schema as well.

Issue #133, opened 2018-11-14
Citation for one DataPaper item doesn't show publisher in DOI Fabrica


kjgarza:
potentially related to datacite/bolognese#36

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_facz)

kjgarza:
test in : https://github.com/datacite/lupo/pull/134

kjgarza:
Citeproc-ruby might not able to convert the citation for `datapaper`

mfenner:
Please look into the citation styles loaded by the respective services. The ones in Lupo are a few months more recent.

Also, I am not sure `bolognese` (and thus the citeproc JSON we generate) understands `datapaper`.

kjgarza:
well `bolognese` is able to format `datapaper` correctly (or with the publisher). It works everywhere (CN, Search, citation formatter), with exception of `lupo` /{doi}?style=**. I'll check sources  anyway.

mfenner:
Citeproc JS in DOI Fabrica:

```
{
  "type": "article-journal",
  "id": "https://doi.org/10.17912/g434-3d85",
  "author": [
    {
      "family": "Bayer",
      "given": "Emily A."
    },
    {
      "family": "Hobert",
      "given": "Oliver"
    }
  ],
  "issued": {
    "date-parts": [
      [
        2018
      ]
    ]
  },
[...]

PR #134, opened 2018-11-14
Publisher in citation does not show on lupo for datapaper type

kjgarza:
related to https://github.com/datacite/lupo/issues/133

DataPaper was added in metadata schema 4.1

the citation doesn't show publishe using the aPI https://api.datacite.org/10.17912/Z4Z9-CE10?style=apa

but using the crosscite website it does

https://citation.crosscite.org/

doesn;t appear in CSL for APA https://github.com/citation-style-language/styles/blob/master/apa.csl

it calls all the way here https://github.com/datacite/bolognese/blob/e0f41e8ce78bccd390678cab07e733efb71d02cd/lib/bolognese/writers/citation_writer.rb#L7

content negotiation also works

https://data.datacite.org/text/x-bibliography;style=apa/10.17912/Z4Z9-CE10

and bolonese as well :

➜ bolognese 10.17912/z4z9-ce10 -t citation

it works with bolognese 1.0.7 and 0.15.7

### DIFF
--- a/spec/fixtures/files/datapaper.xml
+++ b/spec/fixtures/files/datapaper.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<resource
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="http://datacite.org/schema/kernel-4" xsi:schemaLocation="http://datacite.org/schema/kernel-4 http://schema.datacite.org/meta/kernel-4/metadata.xsd">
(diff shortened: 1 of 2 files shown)


mfenner:
@kjgarza what is the purpose of this pull request, as it only contains a test. Is the test failing?

kjgarza:
> I think the issue with not showing the publisher in a formatted citation is caused by the Citeproc. For datasets the publisher field is used, but not for journal articles. We were putting the publisher into the `container-title` field as a workaround.

kjgarza:
this was closed in https://github.com/datacite/lupo/issues/133

Issue #132, opened 2018-11-14
Client appears on /show but not on /index?query:


kjgarza:
very likely the client was created in MDS UI. So its only in the database but not in ES

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_facr)

kjgarza:
![image](https://user-images.githubusercontent.com/1092861/48478523-d40dfd80-e804-11e8-881e-22dd58a599d7.png)

it seems the client was deleted in april

mfenner:
This issue has been resolved.

PR #135, opened 2018-11-14
restrict queries for non-deleted clients only

kjgarza:
related to #132

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -54,6 +54,8 @@ class Client < ActiveRecord::Base
 
   attr_accessor :target_id
 
+  default_scope { where(deleted_at: nil) }
+
   # use different index for testing
   index_name Rails.env.test? ? "clients-test" : "clients"
 


kjgarza:
@mfenner I think looking for individual client is presicily the problem. I was trying to investigate how OSTI.doelabs was able to create DOIs even when their client was deleted. It seems to be that the issue is that current_user at the moment is not filtering deleted clients. For example, the client datacite.test999 is deleted but I can create DOIs from Lupo and Fabrica and reset the password from Fabrica for that account. I will aim to look for another place to set the clause but it seems that definetly has to be related to current_user auth.

mfenner:
I understand that this is about individual clients, so why not put the code in where an individual client is looked up in the controller: https://github.com/datacite/lupo/blob/master/app/controllers/dois_controller.rb#L392-L398

Use 

```
@doi = Doi.where(doi: params[:id]).where(deleted_at: nil).first
```

instead of

```
@doi = Doi.where(doi: params[:id]).first
```

kjgarza:
currently problem with testing with ES.

PR #182, opened 2019-01-07
Xml validation issue649

richardhallett:
This is an attempt to fix datacite/datacite#649.

Goal of XML should not be a required attribute.


mfenner:
Valid XML is required for DOIs in `findable` and `registered` state. This implementation allows for those DOIs to not have any XML.

richardhallett:
Okay I've updated with an explicit test for updates which is the particular use case.
And changed to check for xml presence in validator.

PR #240, opened 2019-04-15
defaults to schema 4version  when the metadata is in schema_org

kjgarza:
https://github.com/datacite/bracco/issues/188


mfenner:
Thanks for spotting this. I would suggest a small change: use schema 4 whenever the input is not in DataCite XML or JSON format. That would also include bibtex, citeproc, ris, etc.

PR #272, opened 2019-05-15
adding delete by prefix rake tasks

kjgarza:
prefix transfers to other RA require a cleaning process. we can use this for that part

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -878,6 +878,15 @@ class Doi < ActiveRecord::Base
     end
   end
 
+  #to be used after DOIS transfer to another RA
+  def self.delete_dois_by_prefix(prefix)
+    collection.where("doi LIKE ?", prefix).find_each do |d|
+      logger = Logger.new(STDOUT)
+      logger.info "Deleted #{d.doi}, last updated #{d.updated.iso8601}."
+      d.destroy
+    end
+  end
+
   # register DOIs in the handle system that have not been registered yet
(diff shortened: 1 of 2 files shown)


app/models/doi.rb
```diff
@@ -878,6 +878,15 @@ def self.delete_test_dois(from_date: nil)
     end
   end
 
+  #to be used after DOIS transfer to another RA
+  def self.delete_dois_by_prefix(prefix)
```
richardhallett: Will the destroy trigger a removal from the ES index?

kjgarza: indexable is included in the DOI model, and it adds a trigger for before destroy 

https://github.com/datacite/lupo/blob/d4b26a9e41e29c926211e54c908b8ba9d6b76342/app/models/concerns/indexable.rb#L16

mfenner:
@kjgarza this is a duplication of this code: https://github.com/datacite/lupo/blob/master/lib/tasks/client.rake#L76-L122

kjgarza:
i think the difference with that is that there is no client for this prefix.  The client must have been deleted in another way. probably before we implemented that other code.

app/models/doi.rb
```diff
@@ -878,6 +878,15 @@ def self.delete_test_dois(from_date: nil)
...
+    collection.where("doi LIKE ?", prefix).find_each do |d|
```
mfenner: This is a very expensive query (using activerecord, using like, and repeating this many times). I suggest using Elasticsearch for this.

Issue #302, opened 2019-07-05
Create an Alias index for Events


kjgarza:
This will allow us to re-index in the background and then switch indexes

So to get to a seamless reindexing you do the following:
https://www.elastic.co/guide/en/elasticsearch/guide/1.x/index-aliases.html

1. create an alias that points to the index with the old mapping (eg. events_alias)*
1. point your application to your alias instead of your index (from events to event_alias, event_alias -> events )*
1. create a new index with the updated mapping (e,g. events_v2)^
1. Stop the EventData Agent (leveriero). To avoid any updates, creates, deletes*
1. move data from old to new index. Use _reindex^
1. atomically move the alias from the old to the new index  (event_alias -> events_v2 )*
1. delete the old stuff (events)

*https://www.rubydoc.info/gems/elasticsearch-api/Elasticsearch/API/Indices/Actions
^https://www.rubydoc.info/gems/elasticsearch-api/Elasticsearch%2FAPI%2FActions:reindex

Some warnings 
> Reindex does not attempt to set up the destination index. It does not copy the settings of the source index. You should set up the destination index prior to running a _reindex action, including setting up mappings, shard counts, replicas, etc.

We could put this in a rake task to execute it later for https://github.com/datacite/lupo/issues/234 .

kjgarza:
We have an alias index already working on Stage which is good for our purposes. We still need to create the alias index in `production`. I won't be able to do this weekend because I have limited internet access.

PR #304, opened 2019-07-05
Feature create alias

kjgarza:
Work to address:  https://github.com/datacite/lupo/issues/302

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -285,7 +285,7 @@ module Indexable
       self.__elasticsearch__.create_index! index: index
       index_name = self.index_name
 
-      client = Elasticsearch::Client.new log: true, host: ENV['ES_HOST']
+      client = Elasticsearch::Model.client
       ### always take origing index even if it was a alias
       index_name = client.indices.get_alias(name: index_name).first.first  if client.indices.exists_alias? name: index_name
       client.reindex body: { source: { index: index_name }, dest: { index: index } }
@@ -294,9 +294,9 @@ module Indexable
     def create_alias(index: nil)
       return nil unless index.present?
 
-      client     = Elasticsearch::Client.new log: true, host: ENV['ES_HOST']
+      client     = Elasticsearch::Model.client
       index_name = self.index_name
-      
+
       client.indices.put_alias index: index, name: index_name
     end
 
@@ -304,7 +304,7 @@ module Indexable
       return nil unless old_index.present? && new_index.present?
       
       # client     = self.gateway.client
-      client = Elasticsearch::Client.new log: true, host: ENV['ES_HOST']
+      client = Elasticsearch::Model.client
       index_name = self.index_name
 
       client.indices.update_aliases body: {
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -296,12 +296,7 @@ module Indexable
 
       client     = Elasticsearch::Client.new log: true, host: ENV['ES_HOST']
       index_name = self.index_name
-
-      client.indices.get_alias(name: index_name).each do |indice|
-        puts "Removing all Aliases first"
-        client.indices.delete_alias index: indice.first, name: index_name
-      end if client.indices.exists_alias? name: index_name
-
+      
       client.indices.put_alias index: index, name: index_name
     end
 
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -280,6 +280,7 @@ module Indexable
     end
 
     def reindex(index: nil)
+      logger = Logger.new(STDOUT)
       return nil unless index.present?
       
       self.__elasticsearch__.create_index! index: index unless self.__elasticsearch__.index_exists? index: index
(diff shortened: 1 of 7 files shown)


app/models/concerns/indexable.rb
```diff
@@ -279,19 +279,34 @@ def count
...
+
     def create_alias(index: nil)
       return nil unless index.present?
 
-      client     = self.gateway.client
+      client     = Elasticsearch::Model.client
       index_name = self.index_name
 
       client.indices.put_alias index: index, name: index_name
+      puts client.indices.get_alias name: index_name
```
mfenner: Why is there a `puts`? Do you want to put this in the log? If so, it may need more context.

app/models/concerns/indexable.rb
```diff
@@ -279,19 +279,34 @@ def count
...
+      return nil unless index.present?
+      
+      self.__elasticsearch__.create_index! index: index unless self.__elasticsearch__.index_exists? index: index
+      index_name = self.index_name
+
+      client = Elasticsearch::Model.client
+      ### always take origing index even if it was a alias
+      index_name = client.indices.get_alias(name: index_name).first.first  if client.indices.exists_alias? name: index_name
+      client.reindex body: { source: { index: index_name }, dest: { index: index } }
+      puts "Index #{index} exists? #{self.__elasticsearch__.index_exists? index: index}"
```
mfenner: See below regarding `puts`.

app/models/concerns/indexable.rb
```diff
@@ -279,19 +279,34 @@ def count
...
+    end
+
     def create_alias(index: nil)
       return nil unless index.present?
 
-      client     = self.gateway.client
+      client     = Elasticsearch::Model.client
```
mfenner: This should be ok, but have you checked whether this works in AWS?

lib/tasks/client.rake
```diff
@@ -14,6 +14,17 @@ namespace :client do
     Client.__elasticsearch__.refresh_index!
   end
 
+  desc "Create Alias"
```
mfenner: I don't understand why we need an alias for the `clients` index? Indexing takes less than 5 min. Is it for zero downtime?

lib/tasks/event.rake
```diff
@@ -1,7 +1,7 @@
 namespace :event do
   desc "Create index for events"
   task :create_index => :environment do
-    Event.__elasticsearch__.create_index!
+    Event.__elasticsearch__.create_index!(index: ENV['INDEX_NAME']||=nil)
```
mfenner: Can you explain this line?

lib/tasks/provider.rake
```diff
@@ -19,4 +19,14 @@ namespace :provider do
   task :refresh_index => :environment do
     Provider.__elasticsearch__.refresh_index!
   end
+
```
mfenner: Again, why do we need aliases for the `providers` index? For zero downtime?

app/models/concerns/indexable.rb
kjgarza: Well yeah, I wanted to show some information about the effects of the changes.  I can add more  context on it.

lib/tasks/client.rake
kjgarza: Yeah, well I thought that once I was at it, I could put it to all indexes.

app/models/concerns/indexable.rb
kjgarza: I have not deployed to Stage yet, so I need to see there. Locally its been working well.

lib/tasks/client.rake
mfenner: That is a funny argument :) Why do we do aliasing and how does it help with `clients` and `providers`?

app/models/concerns/indexable.rb
mfenner: I would use `puts` only in rake tasks or temporarily. Otherwise use the logger.

lib/tasks/event.rake
kjgarza: I just wanted to have a handy option to generate an index with any given name if needed.

For the events index change, I wanted to do the change in two steps.

1. Create a secondary index. (This will help me to do that)
2. Do the reindex.

mfenner: I understand. But how do you remember the index name? And how does a team colleague know that name?

And what does `||= nil` do?

lib/tasks/client.rake
kjgarza: well I mean the change is indeed for aiming to `zero downtime`. But you are right, it would only be needed if the indexes growth considerably.

mfenner: Which means that you remove the rake task or keep it in?

kjgarza: I can remove them

mfenner:
@kjgarza Is there functionality to remove an alias index (not the currently used index)?

kjgarza:
> Some code from this pull request is already in the master branch. Was this intended?

I don't think there is code from this in Master, I doubled check.

> I suggest you implement a naming convention for the alias names and implement a rake task that returns the current index name. That would make it easier to keep track of indexes.

this is a good idea let me look into it 

> Do you plan to delete alias indexes that are no longer used? Do we run the risk of storing lots of obsolete indexes?
> Is there functionality to remove an alias index (not the currently used index)?

Well, I was thinking to use the current delete_index method as I did with create_index, but considering that people might forget the names of the indexes, this might get confusing. So I can look into deleting in another way. I do think that we should put it as a separated step, in case the reindex had fail.

> How do you know that indexing is complete and you can switch to the new alias index?

_reindex returns an array with the full number of documents in the new index. I was thinking to compare that number to the one in the previous index to check.

app/models/concerns/indexable.rb
kjgarza: this has been changed

app/models/concerns/indexable.rb
kjgarza: this has been changed

lib/tasks/provider.rake
kjgarza: this has been changed

kjgarza:
@mfenner I made the change for not requiring index names. Now its all managed internally. I am unsing version for that. Example "events_v1", it works but I can probably do it in a more elegant way, maybe using another gem to manage those numbers as well.

Issue #300, opened 2019-07-01
Including DOIs with events has n+1 problem


mfenner:
DOI metadata are currently fetched individually for every single event.

mfenner:
I suggest trying the `batch-loader` gem, the relevant call has been out into a `BatchLoaderHelper` concern.

Alternatively, fetch DOI metadata in a separate API call, getting the list of DOIs from the Events API call and using `DOI.find_by_id` with the list of these IDs.

kjgarza:
In terms of the performance with batch loading. its 2-3 times faster.
So for the following query for a single DOI events/doi=10.6084/m9.figshare.1407_api_024&include=dois
The response includes 25 events and metadata for 25 DOIs. 

*batch loading: enable*

- 555.4705770045985 ms
- 561.8573510000715 ms
- 581.7548199993325 ms

*batch loading: disabled (DOI.find_by_id )*

- 1741.6881399985868 ms
- 1897.1143399976427 ms
- 1019.5500019981409 ms

Currently, we batch on the `event-id` but we can even improve that if we batch in the `citation-id`. But It will only help when there are inverse-relationships in the response. Nevertheless, the change is rather quick.

mfenner:
Thank you @kjgarza. Can you do more extensive benchmarking, e.g. with `/events`?

kjgarza:
Here its a comparison including three cases: batch-loading enable, batch-loading disabled, and no includes. `No includes` can be seen as some sort baseline as it is time to generate render the events only.


| Batchload enabled                              | No Batchload          | No includes           |
|------------------------------------------------|-----------------------|-----------------------|
| 636.7445210003098 ms                           | 1108.4226549996856 ms | 515.4397269998299 ms  |
| 646.7472290000842 ms                           | 1052.263457999743 ms  | 505.16649900009725 ms |
| 601.9417900001827 ms                           | 1177.9191269997682 ms | 525.0137150001137 ms  |
|                                                |                       |                       |
| /events?doi=10.6084%2Fm9.figshare.1407_api_024 |                       |                       |
| Batchload enabled                              | No Batchload          | No includes           |
| 555.4705770045985 ms                           | 1741.6881399985868 ms | 605.3901689997474 ms  |
| 561.8573510000715 ms                           | 1897.1143399976427 ms | 546.6860540000198 ms  |
| 581.7548199993325 ms                           | 1019.5500019981409 ms | 469.34482200003913 ms |
|                                                |                       |                       |
[...]

Issue #232, opened 2019-04-09
Obtain citation formatted list of DOIs 


kjgarza:
- [x] implementation. 
https://github.com/datacite/lupo/commit/d7cabfcaf94e482779f71e09b6293b06a4c28d29
- [x] review performance with Gbif list: https://api.datacite.org/events?doi=10.15468/dipjcr

We  will use semi “citation format” to show in the citation list. Based o nthe infomration in EventData.

Content Type, Author, YOP, Title, Journal Title, DOI (with http)

mfenner:
Based on today's discussion this is no longer needed.

kjgarza:
- [ ] remove duplicated links if any. The direct and reverse relation could be returned in the events. Implement a filter to avoid double counting.

mfenner:
This is no longer needed.

kjgarza:
I'm not sure if we are referring to the same thing on this ticket. In my mind, this ticket is about formatting the DOI metadata (in whatever serialization one's get it) to the pseudo citation style we defined in the description of this issue. As far as I can see there are 3 serializations we have been using:

A) the /events?include=subj,obj serialization (this was the first implementation we used)
B) the /works serialization (currently I have implemented this ticket using this serialization)
C) the /dois or /events?includes=dois serialization 

If we change to use (C), then we need to redo this ticket.

daslerr:
Formatting is done. We just need to decide how many lists to show.

PR #306, opened 2019-07-07
Feature batch loading and new field for citation_id

kjgarza:
this PR aims to address

-  Including DOIs with events serializer has n+1 problem. We address this problem using batch-loading. https://github.com/datacite/lupo/issues/300
- There can be duplicated citations in the events index. We flag duplicates by adding an id for citations pairs. We create subsequent interfaces to obtain citations filter by such flag. https://github.com/datacite/lupo/issues/232

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -114,7 +114,8 @@ class EventsController < ApplicationController
     dois = total.positive? && params[:extra] ? facet_by_dois(response.response.aggregations.dois.buckets) : nil
     dois_usage = total.positive? && params[:extra] ? facet_by_dois(response.response.aggregations.dois_usage.dois.buckets) : nil
     dois_citations = total.positive? && params[:extra] ? facet_citations_by_year(response.response.aggregations.dois_citations) : nil
-
+    # unique_citations = total.positive? && params[:extra] ? facet_citations_by_dois(response.response.aggregations.unique_citations.dois.buckets) : nil
+ 
     results = response.results
 
     options = {}
@@ -130,7 +131,8 @@ class EventsController < ApplicationController
       registrants: registrants,
       "doisRelationTypes": dois,
       "doisUsageTypes": dois_usage,
-      "doisCitations": dois_citations
+      "doisCitations": dois_citations,
+      # "uniqueCitations": unique_citations
     }.compact
 
     options[:links] = {
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -97,6 +97,7 @@ class EventsController < ApplicationController
                              publication_year: params[:publication_year],
                              occurred_at: params[:occurred_at],
                              year_month: params[:year_month],
+                             unique: params[:unique],
                              page: page,
                              sort: sort)
     end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -126,7 +126,7 @@ class Event < ActiveRecord::Base
     indexes :updated_at,       type: :date
     indexes :indexed_at,       type: :date
     indexes :occurred_at,      type: :date
-    indexes :my_hash,          type: :keyword
+    indexes :citation_id,      type: :keyword
     indexes :cache_key,        type: :keyword
   end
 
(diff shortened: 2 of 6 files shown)


app/models/event.rb
```diff
@@ -159,10 +160,15 @@ def as_indexed_json(options={})
       "updated_at" => updated_at,
       "indexed_at" => indexed_at,
       "occurred_at" => occurred_at,
+      "citation_id" => citation_id,
       "cache_key" => cache_key
     }
   end
 
+  def citation_id
+    [subj_id, obj_id].compact.sort.join("-")
```
mfenner: @kjgarza why do you use `compact`? Both subj_id and obj_id are required.

app/serializers/event_serializer.rb
```diff
@@ -1,3 +1,6 @@
+require 'benchmark'
```
mfenner: Why do you require `benchmark`?

app/controllers/events_controller.rb
```diff
@@ -114,7 +114,8 @@ def index
     dois = total.positive? && params[:extra] ? facet_by_dois(response.response.aggregations.dois.buckets) : nil
     dois_usage = total.positive? && params[:extra] ? facet_by_dois(response.response.aggregations.dois_usage.dois.buckets) : nil
     dois_citations = total.positive? && params[:extra] ? facet_citations_by_year(response.response.aggregations.dois_citations) : nil
-
+    # unique_citations = total.positive? && params[:extra] ? facet_citations_by_dois(response.response.aggregations.unique_citations.dois.buckets) : nil
```
mfenner: Why is this commented out? And what is the relationship between `dois_citations` and `unique_citations`, and do you need both?

kjgarza: Because this aggregation will fail unless the citation_id is already in the index-mapping. So the plan is to add citation_id to the mapping and then enable all the aggregations in a secondary step.  

Do we need `dois_citations` and `unique_citations`? unfortunately kind of yes. The `unique_citations` aggregation is very fast at getting us the total number of unique citations. Unfortunately is a numeric aggregation, hence, we cannot have further sub aggregation; e.g., by year for the chart. That is why we are using `dois_citations` at the moment. 

There seems to be still discussion (robins document) of what criteria will be used to remove duplicated in the chart. But for the moment I will deal with that filtering somewhere else.

app/serializers/event_serializer.rb
kjgarza: leftover from testing will get rid of it

app/models/event.rb
kjgarza: corrrect, addressed

app/controllers/events_controller.rb
mfenner: What sub-aggregations do you need? I thought we said that for aggregations by year, `unique_citations` should give the same numbers.

Aggregations are fast, but we should still be careful that we only do the aggregations we really need.

mfenner:
@kjgarza are there any tests that can be added?

mfenner:
@kjgarza Please describe a description of the feature(s) in the pull request or link to the relevant issue(s).

kjgarza:
I don't think we need to add extra testing for this. The method we use for the unique citations count its literally an ES function, not really code that we wrote. Partially, I think that is usually the reason we don't write tests for aggregations. 

On the batch-loading functionality fall into a similar case, I believe.

app/controllers/events_controller.rb
kjgarza: > What sub-aggregations do you need? I thought we said that for aggregations by year, unique_citations should give the same numbers.

I need a sub-aggregation on the unique items aggregation. 

AS mentioned in the GitHub issue https://github.com/datacite/lupo/issues/232 we obtain the non-duplicated aggregation with the cardinality aggregation metric. Unfortunately, this aggregation doesn't support sub-aggregations. Therefore we cannot get a subsequent aggregation with the time distribution for the chart. So if there are duplicated the chart data would differ from the citations count.

A solution I am working on is to use `collapsing fields`. Collapsing fields filter the list of results but do not affect the aggregations (with ES you cannot have everything :) hehe). One idea using collapsing fields is as follow.  

1. We call API with the cardinality aggregation and the collapsing field filter.
1. We estimate the number of duplicates with the cardinality aggregation. (in the Search Client)
1. If there are duplicates we generate a new yearly citation distribution data-array based on the collapsed results list from the API. This gives us a dataset without duplicates.
1. otherwise we use the default aggregation by year that the API already gives us.

kjgarza: https://github.com/datacite/lupo/pull/306/commits/d454e0d32e8a73319bdff7f69f1db2a5fc1fddbc contains the collapsing fields filter

Issue #307, opened 2019-07-08
Stop using _id for the cursor navigation in our ES searches


richardhallett:
Based on documentation here 
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html

To look into changing the sort that we're using for our search_after cursor.
Currently we use _id but this has potential problems as described in the documentation.

Suggested route is to use
Created timestamp - To ensure order of results
Plus a unique identifier for tie-break resolution. This should have 'doc_values' enabled.

This may be related to the issues under #237

mfenner:
Closed via #308.

PR #308, opened 2019-07-09
Add support for cursor navigation based on created + doi for tie-break

richardhallett:
Related to #307

This means sort orders won't work when cursor specified.

We keep intact existing API that "cursor=1" initiates a cursor query, even though it wont be used in any search results.

### DIFF
--- a/app/controllers/concerns/paginatable.rb
+++ b/app/controllers/concerns/paginatable.rb
@@ -16,6 +16,12 @@ module Paginatable
         page = {}
       end
 
+      # All cursors will need to be decoded from the param
+      if page[:cursor].present?
+        # When we decode and split, we'll always end up with an array
+        page[:cursor] = Base64.strict_decode64(page[:cursor]).split(",")
+      end
+
       if page[:size].present?
         page[:size] = [page[:size].to_i, 10000].min
         max_number = page[:size] > 0 ? 10000/page[:size] : 1
--- a/app/controllers/concerns/paginatable.rb
+++ b/app/controllers/concerns/paginatable.rb
@@ -22,7 +22,8 @@ module Paginatable
           # When we decode and split, we'll always end up with an array
           page[:cursor] = Base64.strict_decode64(page[:cursor]).split(",")
         rescue ArgumentError
-          raise "Invalid base64 used for cursor pased from query params"
+          # If we fail to decode we'll just default back to an empty cursor
+          page[:cursor] = []
         end
       end
 
--- a/app/controllers/concerns/paginatable.rb
+++ b/app/controllers/concerns/paginatable.rb
@@ -18,8 +18,12 @@ module Paginatable
 
       # All cursors will need to be decoded from the param
       if page[:cursor].present?
-        # When we decode and split, we'll always end up with an array
-        page[:cursor] = Base64.strict_decode64(page[:cursor]).split(",")
+        begin
+          # When we decode and split, we'll always end up with an array
+          page[:cursor] = Base64.strict_decode64(page[:cursor]).split(",")
+        rescue ArgumentError
+          raise "Invalid base64 used for cursor pased from query params"
+        end
       end
 
       if page[:size].present?
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -31,6 +31,16 @@ class DoisController < ApplicationController
 
     page = page_from_params(params)
 
+    # For DOI's we need to decode the page cursor
+    if page[:cursor].present?
+      # Cursor navigation overrides any sort orders specified.
+      sort = { created: { order: 'asc' }, doi: "asc"}
+      page[:cursor] = Base64.decode64(page[:cursor]).split(",")
+      show_cursor_link = true
+    else
+      show_cursor_link = false
+    end
+
     sample_group_field = case params[:sample_group]
       when "client" then "client_id"
       when "data-center" then "client_id"
@@ -155,7 +165,7 @@ class DoisController < ApplicationController
           options[:meta] = {
             total: total,
             "totalPages" => total_pages,
-            page: page[:cursor].blank? && page[:number].present? ? page[:number] : nil,
+            page: show_cursor_link == false && page[:number].present? ? page[:number] : nil,
             states: states,
             "resourceTypes" => resource_types,
             created: created,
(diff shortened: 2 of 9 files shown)


app/controllers/dois_controller.rb
```diff
@@ -31,6 +31,16 @@ def index
 
     page = page_from_params(params)
 
+    # For DOI's we need to decode the page cursor
+    if page[:cursor].present?
```
mfenner: I would implement this here: https://github.com/datacite/lupo/blob/master/app/controllers/concerns/paginatable.rb#L6-L29

app/controllers/dois_controller.rb
```diff
@@ -180,8 +190,9 @@ def index
               query: params[:query],
               "provider-id" => params[:provider_id],
               "client-id" => params[:client_id],
-              "page[cursor]" => page[:cursor].present? ? Array.wrap(results.to_a.last[:sort]).first : nil,
-              "page[number]" => page[:cursor].blank? && page[:number].present? ? page[:number] + 1 : nil,
+              # The cursor link should be based on the ES sort, but we want to encode it into a single string
+              "page[cursor]" => show_cursor_link == true ? Base64.encode64(Array.wrap(results.to_a.last[:sort]).join(',')) : nil,
```
mfenner: Consider `strict_encode64` to not add linefeeds.

app/models/concerns/indexable.rb
```diff
@@ -104,8 +104,8 @@ def query(query, options={})
       # enable cursor-based pagination for DOIs
       if self.name == "Doi" && options.dig(:page, :cursor).present?
         from = 0
-        search_after = [options.dig(:page, :cursor)]
-        sort = [{ _id: { order: 'asc' }}]
+        search_after = Array.wrap(options.dig(:page, :cursor))
```
mfenner: Consider doing the base64decode here.

mfenner:
Consider adding a test.

app/controllers/dois_controller.rb
richardhallett: I didn't do this because paginatable is something that applies across everything that implements it. I kept it as being something that just returns the cursor from the paras into the "page", it doesn't therefore know it's a token or anything, that's up to individual controllers to decide if it it needs to do additional work. This is also why I force a sort here as this is specific to dois handling.

app/controllers/dois_controller.rb
richardhallett: Sounds reasonable, wasn't aware of the difference in Ruby. Will add.

app/models/concerns/indexable.rb
richardhallett: I did originally put the decode here, but I felt that the query here doesn't really know it's DOI's and I feel like it makes it more in-line with the other cursors if it just takes what is being told is the cursor.
Additionally one of the reasons it doesn't work so well is that we have to support the cursor of page[cursor]=1 - With the logic handling in dois, it actually means that the page[cursor] gets set to an empty [] due to 1 not decoding properly, this then means it wont try and do this logic.

I could perhaps try and re-work it so we do an additional check here.

app/controllers/dois_controller.rb
mfenner: Don't we want to support cursor pagination everywhere? Or do you think it is really only needed for the large indexes `dois` and `events`? In the latter case I agree with your approach.

app/models/concerns/indexable.rb
mfenner: Makes sense. 

We can change the logic of when cursor-based pagination is triggered, doesn't have to be that a page[cursor] parameter is present. And we might need some error handling for cursors that don't make sense, including `1`.

app/models/concerns/indexable.rb
```diff
@@ -101,30 +101,31 @@ def query(query, options={})
...
         sort = [{ created: "asc", doi: "asc"}]
-      elsif self.name == "Event" && options.dig(:page, :cursor).present?
-        from = 0
-        search_after = [options.dig(:page, :cursor)]
+      elsif self.name == "Event" && !options.dig(:page, :cursor).nil?
         sort = [{ _id: { order: 'asc' }}]
-      elsif self.name == "Activity" && options.dig(:page, :cursor).present?
-        from = 0
-        search_after = [options.dig(:page, :cursor)]
+      elsif self.name == "Activity" && !options.dig(:page, :cursor).nil?
```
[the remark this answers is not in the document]
richardhallett: Yes, because if there is any kind of cursor it should always change the default sort.
Arguable none of this code should be in here and instead just be passed as cursor sort options perhaps, but I think this is an acceptable compromise without refactoring more parts.
Additionally cursor should always be [] not "", I could ensure this with an error check at the start perhaps.

app/controllers/concerns/paginatable.rb
```diff
@@ -16,6 +16,12 @@ def page_from_params(params)
         page = {}
       end
 
+      # All cursors will need to be decoded from the param
+      if page[:cursor].present?
+        # When we decode and split, we'll always end up with an array
+        page[:cursor] = Base64.strict_decode64(page[:cursor]).split(",")
```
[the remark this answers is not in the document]
richardhallett: Fair enough, this is in part originally why I had this just in each of the controllers I'm still not sure it shouldn't be there. Because arguably it should always be a base64 encoded string if it's coming in from the query string
I guess we can do some safety checks though to ensure we catch potential problems.

app/models/concerns/indexable.rb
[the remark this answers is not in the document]
mfenner: you can force an array with `Array.wrap(var)`. And `elsif self.name == "Activity" && options.dig(:page, :cursor)` should be the same as `elsif self.name == "Activity" && !options.dig(:page, :cursor)nil?`

app/controllers/concerns/paginatable.rb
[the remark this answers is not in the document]
mfenner: I would add a `begin/rescue` block, catching an error if the cursor is not properly base64 encoded. The error you need to catch is `ArgumentError` (with message "invalid base64").

app/controllers/concerns/paginatable.rb
```diff
@@ -16,6 +16,17 @@ def page_from_params(params)
...
       end
 
+      # All cursors will need to be decoded from the param
+      if page[:cursor].present?
+        begin
+          # When we decode and split, we'll always end up with an array
+          page[:cursor] = Base64.strict_decode64(page[:cursor]).split(",")
+        rescue ArgumentError
+          # If we fail to decode we'll just default back to an empty cursor
+          page[:cursor] = []
```
richardhallett: Note we don't want to throw an error back here because cursor=1 is allowed in our API and we need to kick start cursors sometimes.

mfenner: Looks good.

mfenner:
@richardhallett tiny comment: use `cursor_values` instead of `cursorValues`. More idiomatic.

PR #311, opened 2019-07-15
adding filtered aggregations

kjgarza:
We are adding more and more aggregations to our endpoints. They have an impact in performance. We can make use of aggregation selector for specific use cases. For example, when we need them in Fabrica or in Search for citations

for example:

`/events` will give you all the normal aggregation we have

`/events?aggregations=metrics_aggregations` will give you the metrics aggregation but exclude the normal ones

`/events?aggregations=metrics_aggregations,query_aggregations` will give you both the metrics aggregations and the normal aggregations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 rvm:
 - 2.4.4
 sudo: required
-dist: trusty
+dist: xenial
 
 services:
   - mysql
@@ -12,13 +12,6 @@ services:
 addons:
   code_climate:
     repo_token: $CODECLIMATE_REPO_TOKEN
-  apt:
-    packages:
-      - oracle-java9-set-default
-      - mysql-server
-      - mysql-client
-    sources:
-      - mysql-5.7-trusty
 
 before_install:
   - curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.3.2.deb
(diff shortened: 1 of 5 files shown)


mfenner:
@kjgarza while I like the general idea, I am not sure it really solves a problem. Is there a significant impact on performance with the current aggregations? My understanding is that the most time is spent generating the JSON in the serializer rather than the aggregations. There might also be potential for optimization in the `facetable` concern.

The downside of your approach is that it adds complexity, and we should avoid that if possible.

kjgarza:
I don't think the current aggregations have a big issue but we are adding more for specific things (usage [the ones we will use in the widget] and citations, which need filtered-scripts) that overload the normal calls. I think separating them would be useful in those terms plus it would help in the development side to avoid conflicts. The later, I think is one of the main advatages

mfenner:
Let's discuss this in our next Kanban meeting.

kjgarza:
Also, last week, on adding the citation_year changes also include this selector. Because i thought it was  no brainer and just needed to need ou time to merge it https://github.com/datacite/lupo/commit/f3c1a3228f811064603255a2c0215c00ef1635e0

kjgarza:
implemented in https://github.com/datacite/lupo/commit/f3c1a3228f811064603255a2c0215c00ef1635e0

PR #336, opened 2019-08-09
Check for title in DOI creation

kjgarza:
reverse this change https://github.com/datacite/lupo/commit/3f743f4ab95d00bc22383ad9f151aa9f2730e574#diff-bfc22500c2c731bcf2d6fa8f1e22a6dd

as it still needs a presence check. when you have an empty string or array. In those cases, #presence returns nil.

I added validation for titles anyway

Fixes : https://github.com/datacite/datacite/issues/811

### DIFF
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -621,7 +621,7 @@ class DoisController < ApplicationController
     # merge attributes from xml into regular attributes
     # make sure we don't accidentally set any attributes to nil
     read_attrs_keys.each do |attr|
-      p.merge!(attr.to_s.underscore => p[attr] || meta[attr.to_s.underscore] || p[attr]) if p.has_key?(attr) || meta.has_key?(attr.to_s.underscore)
+      p.merge!(attr.to_s.underscore => p[attr].presence || meta[attr.to_s.underscore] || p[attr]) if p.has_key?(attr) || meta.has_key?(attr.to_s.underscore)
     end
     p.merge!(version_info: p[:version] || meta["version_info"]) if p.has_key?(:version_info) || meta["version_info"].present?
 
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -621,7 +621,7 @@ class DoisController < ApplicationController
     # merge attributes from xml into regular attributes
     # make sure we don't accidentally set any attributes to nil
     read_attrs_keys.each do |attr|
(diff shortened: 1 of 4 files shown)


mfenner:
The validation for the presence of `titles` is required. The `check_titles` validation is triggered only if there is content in `titles`.

mfenner:
How are these changes fixing datacite/datacite#811 ?

kjgarza:
> The validation for the presence of titles is required. The check_titles validation is triggered only if there is content in titles.

there is already a validation for titles in metadata already in place. I first added because for the purpose of this error I thought it was useful. However, there is already a check somewhere else.

kjgarza:
> How are these changes fixing datacite/datacite#811 ?

currently, DOIs updates are removing the `titles`. Previously if the title information was not part of the JSON object it would default to the :xml/meta object in :

```p.merge!(attr.to_s.underscore => p[attr].presence || meta[attr.to_s.underscore].presence``` 

However since the update (`enforce type for doi json fields.`) It would not default to the :xml/meta object anymore. Instead, it would write the empty array that the JSON object has. 

I think the current logic in `safe_params` assumes that JSON object values have precedence over whatever is in the :xml/meta-object. However, I think that might not be the case for updates, particularly those coming from the Fabrica Upload Form.

mfenner:
You are right, the validation of the presence of titles happens in the `xml_schema` validator.

I still don't see how your changes allow the update of optional attributes (e.g. `descriptions`) to `[]`.

mfenner:
And please keep in mind that there is a second round of checks: `p[attr]) if p.has_key?(attr) || meta.has_key?(attr.to_s.underscore)`.

mfenner:
@kjgarza please provide one or more tests that address the issue. The example spec you provide shows the expected behaviour before your changes in the `dois_controller`, as JSON attributes always overwrite what is provided via the `xml`attribute. There isn't really a use case for providing the same attribute both ways, e.g. Fabrica does either one or the other, but never both.

I made some minor changes.

kjgarza:
Screenshot 2019-08-13 at 13 17 16


Well the problem is happening on DOI update using `file upload` and in those cases Bracco/Fabrica its sending titles as an empty array "[ ]"  in the JSON object as per  

https://github.com/datacite/bracco/blob/b87f8549d56295b95a18964b2404609eaa760955/app/controllers/clients/show/dois/show/modify.js#L29 

So, yeah like you said, If JSON object values have precedence over whatever is in the :xml/meta-object, then we need to make a change in BRacco/fabrica.

PR #339, opened 2019-08-27
Feature get metrics from GraphQL API

kjgarza:
Define aggregations for citations, views and downloads  as well as crossref and datacite citations
Implement Types for  metrics
Define JSON Response for metrics response

addresses https://github.com/datacite/spitz/issues/2

image (11)


mfenner:
Please change description for `viewCount` and `downloadCount` to say "The count of DOI views according to the COUNTER Code of Practice" instead of "The count of DOI views according to the COUNTER code of Practice for Research Data".

Issue #362, opened 2019-11-25
refractor citations retrieval for multiple `dois` in the API


kjgarza:
related https://github.com/crosscite/doi-metadata-search/issues/156

mfenner:
Please add epic and estimate.

PR #364, opened 2019-11-25
Refractor multiple dois in events

kjgarza:
@mfenner  I want to pick up your brain in the architecture change. So basically I am using now a query object. I'm following the same principle as I was doing it with the widget. The difference is that there is script filtering and I am using a query object.

I need to clean it up, add more tests and check some interfaces, but the idea is here and it works so far.

related: https://github.com/datacite/lupo/issues/362


app/queries/events_query.rb
```diff
@@ -0,0 +1,39 @@
...
+    "references"
+  ]
+
+  PASSIVE_RELATION_TYPES = [
+     "is-cited-by",
+     "is-supplemented-by",
+     "is-referenced-by"
+  ]
+
+  def initialize(relation = Event.new)
```
mfenner: I don't understand this line.

app/queries/events_query.rb
```diff
@@ -0,0 +1,39 @@
+class EventsQuery
```
mfenner: Why is this in a `queries` subfolder?

app/queries/events_query.rb
mfenner: Why is EventQuery initialized with a single relation?

app/queries/events_query.rb
```diff
@@ -0,0 +1,39 @@
...
+    results.any? ? results.first.total.value : 0
+  end
+
+  def citations(doi)
+    doi.downcase.split(",").map do |item|
+      {id: item, count: EventsQuery.new.doi_citations(item)}
+    end
+  end
+
+
```
mfenner: Whitespace...  Look at rubocop linting.

spec/queries/events_query_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+describe EventsQuery, elasticsearch: true do
+  # before(:each) { allow(Time.zone).to receive(:now).and_return(Time.mktime(2015, 4, 8)) }
+
+  context "citation events" do
+    # subject { create(:event_for_datacite_related) }
+    let!(:event) { create(:event_for_datacite_related,  subj_id:"http://doi.org/10.0260/co.2004960.v2", obj_id:"http://doi.org/10.0260/co.2004960.v1") }
+    let!(:event) { create_list(:event_for_datacite_related, 3,obj_id:"10.5061/dryad.47sd5/2", relation_type_id: "references") }
```
mfenner: This is not going to work, the second event overwrites the first one.

app/queries/events_query.rb
kjgarza: Mostly trying to follow query object pattern here. 

https://medium.flatstack.com/query-object-in-ruby-on-rails-56ea434365f0

Both the controller and the model seem very crowded and found it better here.

app/queries/events_query.rb
kjgarza: I need to set up that. I haven't. right now I'm more interested on the feedback of the architecture. That's why i sent it earlier.

kjgarza:
> I am worried that the architecture for aggregating citations, references and relations has become rather messy, and put a load on the API.

The change doesn't follow the previous architecture anymore. `Citations` (not references, or relations) are completely obtained separated from the rest. I see a path to re-architect references and relations in the same way but I am not doing that now(not a priority atm). For example, the `compilations within, max: [75/5m]` errors shouldn't be shown anymore, that is for sure.

https://sentry.io/organizations/datacite/issues/1346228335/?query=is%3Aunresolved

What would help to provide context?

Pushing to test might help to see it working but as with the rest of the citations. It's difficult to assess in other places rather than `production`.

app/queries/events_query.rb
mfenner: Ok, but the example is about ActiveRecord queries not Elasticsearch. `relation` has a very different meaning there.

app/queries/events_query.rb
mfenner: Please take care of well-formatted code from the start if you want me to accept your pull-requests.

Issue #360, opened 2019-11-22
 There used to be views and downloads on this page. 


kjgarza:
Currently the API is returning 0 views and 0 downloads. It also happens when calling graphql. Example:
```
{
  dataset(id:“10.7272/q6g15xs4”) {
    id
    titles{
      title
    }
    viewCount
    downloadCount
    citationCount
  }
}
```
This might be related to the list of breaking changes from `ElasticSearch@7`: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#breaking_70_scripting_changes
For example
> getMillis() should be replaced with toInstant().toEpochMilli()
Its used in aggregations.

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_146tf)

kjgarza:
The problem is a bit more complex but not that bad.

The reason why we get 0 usage and citations is that there are actually 0 events that follow the filtering criteria. 

The criteria is not wrong but the wya EventData is indexing events is. So all the index mapping is done with `camelCase` however all event proporties are converted to kebab case in the controller.

here:
https://github.com/datacite/lupo/blob/d875b92346f24361d95e8ab445a326ab83dc254f/app/controllers/events_controller.rb#L237

On can see that in kibana how the values in the index are in the kebaba case version of `date_published` but not in `datePublished` which is the one in the ES mapping.
https://8cf39df5abd84a29b63175bdce2ba1c8.eu-west-1.aws.found.io/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-5h,to:now))&_a=(columns:!(obj.date_published,source_id,obj.id),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'9b656b20-0ea1-11ea-99a9-b129f692045f',key:source_id,negate:!f,params:(query:datacite-usage),type:phrase,value:datacite-usage),query:(match:(source_id:(query:datacite-usage,type:phrase))))),index:'9b656b20-0ea1-11ea-99a9-b129f692045f',interval:m,query:(language:kuery,query:''),sort:!(!(updated_at,desc),!(source_id,desc)))

![image](https://user-images.githubusercontent.com/1092861/69738939-229fe880-1137-11ea-9ab5-09f6d6b0d903.png)




[...]

Issue #352, opened 2019-11-08
Querying by subj or obj internal attributes doesn't work


kjgarza:
## problem

One cannot query for a list, this should give us all the events with crossref in both nodes

`/events?query=obj.registrantId%3Acrossref*%20AND%20subj.registrantId%3Acrossref*`

# hypothesis

Most of the indexed information includes `obj.registrant_id`  the schema indicates that i should be `obj.registrantId` 

https://github.com/datacite/lupo/commit/ff565d849c8e9083e36f5b4d1337c57ad7a80a5b
in the database they are all beeing save as registrant_id. But levreiro is sending registrantId. So there might be a conversion.

Potential solutions to make the query

- change the whole mapping and reindex all the events (it doesn need to reimport) <- because in the database everythin is saved with snacke_case

PR #365, opened 2019-11-27
Bug incorrect ES mapping

kjgarza:
to address https://github.com/datacite/lupo/issues/360 and https://github.com/datacite/lupo/issues/352

Most of the indexed information includes obj.registrant_id the schema indicates that i should be obj.registrantId

ff565d8
in the database they are all beeing save as registrant_id. But levreiro is sending registrantId. So there might be a conversion.

Potential solutions to make the query

change the whole mapping and reindex all the events (it doesn't need to reimport) <- because in the database everything is saved with snacke_case

we might need to do make the change over the weekend as we will need to reindex the events.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Event < ActiveRecord::Base
   # include helper module for query caching
   include Cacheable
@@ -23,8 +25,8 @@ class Event < ActiveRecord::Base
   # include state machine
   include AASM
 
-  aasm :whiny_transitions => false do
-    state :waiting, :initial => true
+  aasm whiny_transitions: false do
+    state :waiting, initial: true
     state :working, :failed, :done
 
     # Reset after failure
@@ -67,7 +69,7 @@ class Event < ActiveRecord::Base
   ]
 
   ACTIVE_RELATION_TYPES = [
-    "cites", 
+    "cites",
     "is-supplement-to",
     "references"
   ]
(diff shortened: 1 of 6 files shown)


mfenner:
We have seen issues with snake-case vs. camelCase in other places and need to be more consistent about this.

kjgarza:
indeed, the main issue is this line

https://github.com/datacite/lupo/blob/d875b92346f24361d95e8ab445a326ab83dc254f/app/controllers/events_controller.rb#L237

it modifies all attributes. So I guess we didn't see the implications of it.

app/models/event.rb
```diff
@@ -117,18 +119,18 @@ class Event < ActiveRecord::Base
       type: { type: :keyword },
       id: { type: :keyword },
       uid: { type: :keyword },
-      proxyIdentifiers: { type: :keyword },
-      datePublished: { type: :date, format: "date_optional_time||yyyy-MM-dd||yyyy-MM||yyyy", ignore_malformed: true },
```
kjgarza: @mfenner these are the attributes we cannot access because of the casing. 

If we do not change the mapping, we could:

A) modify all the values in the database to camelCase (either re importing all events or transforming all of them)
B) change the indexJob so the transformation happens there during every index.

mfenner: I think we need both.

kjgarza:
discarded strategy

digitaldogsbody:
Hi @kjgarza, since this strategy has been discarded, is it ok for me to remove the stale branch associated with this PR?

kjgarza:
SGTM

PR #369, opened 2019-12-03
pre-index linking data

kjgarza:
addresses https://github.com/datacite/spitz/issues/6

According to our criteria, Every Event can be at most two `types of links` at the same time:

- An event of the type `citation-link` will always be also  `reference-link` at the same time. A `citation-link` for the subjectt of the event, it's a  `reference-link` for the object and vice-versa.
- An event of the type  `relation-link` can only be a  `reference-link` type only.
- An event of the type  `view-link` can only be a  `view-link` type only.
- An event of the type  `download-link` can only be a  `download-link` type only.

link_types for 10.14454/9ofdqasd3
`10.14454/9ofdqasd3-citation`
`10.14454/9ofdqasd3-reference`
`10.14454/9ofdqasd3-relation`
`10.14454/9ofdqasd3-view`
`10.14454/9ofdqasd3-download`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,4 @@
-require: rubocop-performance
+# require: rubocop-performance
 inherit_from: .rubocop_todo.yml
 
 # from https://github.com/rails/rails/blob/master/.rubocop.yml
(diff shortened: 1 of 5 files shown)


mfenner:
Closing this pull request for now, first needing an architecture discussion.

PR #394, opened 2020-01-21
reduce number ElasticSearch queries for metrics

kjgarza:
We are making too many ElasticSearch queries to obtain metrics. We are aiming to reduce that for the researcher profile implementation.

**Solution**
Rather than putting the response metrics inside the DOI object, we can put all the metrics in an array in the response separated. So the response would look like in the figure below. Then in the frontend interface (ember.js) we can mix in the metrics into every doi display.

carbon (16)

**How many `queries` do we at least need to do?**
- For views and downloads, we can have 1 `query` and multiple aggregations. 
- For citations, we need at least 2 `queries` and then use ruby logic to merge the responses of the queries. We need to queries because of the criteria for citations. You need two lists of conditions. The only way to reduce it to one query is if we know which events, for each doi, are citations (as mentioned in https://github.com/datacite/lupo/pull/369).

### DIFF
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -246,13 +246,24 @@ class DoisController < ApplicationController
 
             # sparse fieldsets
             fields = fields_from_params(params)
+            dois_serialized = nil
             bm = Benchmark.ms {
               if fields
-                render json: DoiSerializer.new(results, options.merge(fields: fields)).serialized_json, status: :ok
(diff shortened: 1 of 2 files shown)


digitaldogsbody:
Hi @kjgarza, I am in the process of tidying up some of the stale branches in this repo - can the associated branch for this PR now safely be pruned? It appears that the changes are mostly to files that no longer exist in the current repo.

kjgarza:
SGTM

PR #396, opened 2020-01-21
reduce number of ES queries to obtain metrics in DOI list for researcher profile

kjgarza:
## Description

Optimized the number of ElasticSearch queries to obtain metrics. 
To avoid 1+N request problems I modified the queries for metrics to request metrics for groups of dois. 
We use very small aggregation per query and the overhead is mostly in the citations query which requires an extra query due to the citations criteria conditions.

related to https://github.com/datacite/bracco/issues/270

## Status
- [x] Ready for Review

## Types of changes

## Non Functional Requirement
- [x] Tests Cover Changes
- [ ] All new and existing tests passed.
- [ ] Documentation


kjgarza:
Some unwanted database changes went through. I'll remove them

app/controllers/concerns/metrics_helper.rb
```diff
@@ -1,17 +1,45 @@
...
-      EventsQuery.new.doi_citations(doi)
+  included do
+
+    def get_metrics_array(dois)
+
+      citations   = EventsQuery.new.citations(dois)
+      views       = EventsQuery.new.views(dois)
+      downloads   = EventsQuery.new.downloads(dois)
+
+      first_merge = merge_array_hashes(citations, views)
```
mfenner: can this not be done in one step?

app/controllers/concerns/metrics_helper.rb
```diff
@@ -1,17 +1,45 @@
 module MetricsHelper
   extend ActiveSupport::Concern
 
-  class_methods do
-    def doi_citations(doi)
-      EventsQuery.new.doi_citations(doi)
+  included do
+
+    def get_metrics_array(dois)
+
```
mfenner: linting...

kjgarza:
> I would merge the metrics with the rest of the metadata in the controller, and not the serializer.

@mfenner the problem with that is that all the other metadata comes in a ES:results object merging it with the metrics data would require to transform it into a hash, then that would need to mean we would need to modify the whole Serializer to expect hashes instead of objects like we normally do. Mixing it in the serializer seems like a normal approach https://github.com/Netflix/fast_jsonapi#using-helper-methods

mfenner:
Regarding whether to merge the metrics in the controller or serializer, think about what kind of information it is, for display only (serializer), or more fundamental and maybe also used in other ways. This could for example be benchmarking.

It is easy to generate objects from hashes, e.g. with `OpenStruct`, `Struct` or `Mash` (in the `hashie` gem).

app/controllers/concerns/metrics_helper.rb
kjgarza: changed to one step

app/controllers/concerns/metrics_helper.rb
kjgarza: changed already

Issue #351, opened 2019-11-08
To remove Events with the source DataCite-Crossref with Crossref Dois in both nodes


kjgarza:
### problem

There are Events with the source DataCite-Crossref with Crossref Dois in both nodes. Example:

https://api.datacite.org/events/747cbf1f-ac7a-4de8-beba-f459bf0114f7

https://api.datacite.org/events?subj-id=https%3A%2F%2Fdoi.org%2F10.1017%2Fs1041610218001187

### hypothesis

The subj of the event https://doi.org/10.1017/s1041610218001187  is a Crossref.citation DOI. This might indicate that crossref.citations dois were imported into event data and all their related identifiers were converted into more events. 

This is most likely to have happened during a rake import, as the normal workflow excludes crossref.citations but rake tasks do not. 

We need to access the magnitud of this.

kjgarza:
Events with the 

So we have been running the script for a while. So far I have captured 744,277 events with problem. I have been flagging them.Now identify them all by ID. the label is `crossref_citations_error` in the `state_event` field

List of events (we can paginate through them and get all of the ids):
https://api.datacite.org/events?aggregations=state_aggregations&source=datacite-crossref&query=state_event%3Acrossref_citations_error

To see the summary only (to see the extent of the damage)
https://api.datacite.org/events?aggregations=state_aggregations&source=datacite-crossref&query=state_event%3Acrossref_citations_error&page%5Bsize%5D=0

kjgarza:
@mfenner Crossref still needs access to these events while they delete their copies.

However,  I am wondering whether we actually want to delete all the events with an errors of this type or if we prefer just to hide them away? Currently all of them are labeled but they are still in the database/index. 

I guess one argument is that we just keep them there as evidence in case we need to explain later. What do you think?

mfenner:
The events database and index are already pretty big, so I would avoid hiding instead of deleting if at all possible.

kjgarza:
Events have been filtered and there is a task to remove it. 
We need to wait from Crossref Confirmation to remove the events. 
We change this from `bug` to `housekeeping`

PR #398, opened 2020-01-23
rake task to get list and totals of event with corssref events in bot……h nodes.

kjgarza:
## Description
rake tasks to assess the magnitude of events that have crossref-to-crossref relationships, which were wrongly indexed. See https://github.com/datacite/lupo/issues/351 for more details.

This rake task obtains the list of all the events and saves it into a gist file for quick access.

Criteria or from events is:
subj_id in datacite-crossref,datacite-related events MUST always be a DataCite DOI

ITs a new rake task that triggers a method in the events model

## Status
- [x] Ready for Review

## Types of changes

## Non Functional Requirement
- [x] Tests Cover Changes 
- [ ] ~~Documentation~~


lib/tasks/event.rake
```diff
@@ -71,6 +71,15 @@ namespace :crossref do
   end
 end
 
+namespace :subj_id_check do
+  desc 'checks datacite-crossref,datacite-related events have a DataCite DOI in the subject node'
```
mfenner: Please use shorter `desc`

lib/tasks/event.rake
```diff
@@ -71,6 +71,15 @@ namespace :crossref do
...
+  task :check => :environment do
+    cursor = ENV['CURSOR'].to_s.split(",") || [Event.minimum(:id),Event.minimum(:id)]
```
mfenner: Please align with how we use `cursor` for events in other places, or use a different term.

app/models/event.rb
```diff
@@ -489,6 +489,37 @@ def access_method
...
+        break unless response.results.results.length > 0
+
+        logger.info "[DoubleCheck] DoubleCheck #{response.results.results.length}  events starting with _id #{response.results.to_a.first[:_id]}."
+        cursor = response.results.to_a.last[:sort]
+
+        # dois = response.results.results.map(&:subj_id)
+        events = response.results.results
+        events.lazy.each do | event|
+          subj_prefix = event.subj_id[/(10\.\d{4,5})/,1]
+          File.open("evens_with_double_crossref_dois.txt", "a+") do |f|
```
mfenner: Please don't use `File.open` in a container, as file storage is not persistent. Rather write to another place, e.g. via API call.

app/models/event.rb
```diff
@@ -489,6 +489,37 @@ def access_method
...
+            f.write(event.uuid, "\n")
+            total_errors= total_errors+1
```
mfenner: linting...

lib/tasks/event.rake
kjgarza: done

app/models/event.rb
kjgarza: done

app/models/event.rb
kjgarza: Now i pushing the data via API to storage outside the container. However, I am still creating a file to make the whole process faster and send it over a single API call.

kjgarza:
@mfenner Crossref needs a list of the events that have problems to deal with them on their side (flag/erase/hide IDK). The file is the best way to get them the list. Saving the file to a gist it's a fast, simple, and straightforward way to share the list with crossref. The rake task only needs to use this time. 

I could write this rake task into a script and run it from my computer. But the rake task has the advantage of being a speedy way to deliver this to crossref and its easy to manage.

This rake task doesn't have to live permanently in our code. We can remove our repurpose later. We have done that in the past in with, for example, `migrate_landing_page`. See https://github.com/datacite/lupo/blob/ca71f26d5653ae913ae45678349ee2e933e296d9/app/models/doi.rb#L1532

mfenner:
I am not completely convinced, but please move ahead, using gists.

kjgarza:
thanks

PR #406, opened 2020-01-30
Feature seeds for validating in our Development or Stage environment

kjgarza:
We often do not have good dummy data in our development or staging environments to validate features. We often rely on copying data from the production of generating data manually. This takes too much time and is not reproducible. 

We need a way to make this in a more efficient way.

Potential use case:

- **As a** developer **I want** an easy way to get LUPO running locally with data **So that** so that I can validate my changes

- **As a** developer **I want** an easy way to create useful dummy data in our Staging/Deve enviroment **So that** so that I can validate my changes

## Technical proposal

To use Rails `seed`  files in combination with `faker` and `Factory_bot` gems to generate data that we need in a consistent fashion.

We can use the `seedbank` gem to generate a folder structure for the different use cases we will need in the future.

it would be as easier as:

> bundle exec rake db:seed:development

if you were to use it locally, it might need to do

> bundle exec rake db:create
> bundle exec rake db:schema:load
> bundle exec rake db:seed:development

### Backend

- [x] [Generate data necessary for the Researcher Profile]()
- [x] [include safeguard for the tasks to be only run in the stage enviroment]()


mfenner:
Please clarify whether the seed data are for `test`, as the title implies, or for `development` and `stage`.

spec/factories/default.rb
```diff
@@ -1,8 +1,8 @@
-require 'faker'
+require "faker"
 
 FactoryBot.define do
   factory :user do
-    sequence(:name) { |n| "Josiah Carberry{n}" }
+    sequence(:name) { |_n| "Josiah Carberry{n}" }
```
richardhallett: I dont think this will work now, as it's _n not n, so it'll just end up printing probably 0 or whatever defaults

kjgarza:
>Seedbank has not been updated to work with Rails 5. I've used it with 5.x apps and am working on a new version specifically for 5.x. Other people are also reporting using it with no problems.

There are comments from maintainer on that and it works as described anyway.

With regards to using factory_bot, basically it used to create data in the database, in principle if the factories are good enough for automated testing, they would be good enough create consistent data for validation.

mfenner:
I am not sure I am convinced regarding `factory_bot`. The requirements for a system used in test are different. Have you looked into other approaches, including a simple SQL dump?

kjgarza:
Comparison with other approaches


| Approach            | File Management | Data Generation | Adv | DisAdv |
|---------------------|-----------------|-----------------|-----------------|-----------------|
| Proposal A  (this PR)        | seedbank gem    | factoryBot Gem  | "Its maintained with rspec tests. Since you already have several factories used for testing, that you already keep up to date it makes a lot of sense to reuse them instead of rewriting that logic. One can create multimes records with a one-liner. Validation-data-as-code"
| Usign SQL files     | seedbank gem    | SQL data dump   | | "Need to create SQl datadump for specific use cases. Old SQL data dumps need to be updated when there are changes in the model and schema (on migrations)."|
| Usign Rails Models  | seedbank gem    | Using Ruby      | Validation-data-as-code | "One will need to maintain the code in the seeds files after every code change. One needs to populate all nedded fields or write logic to create values."|
| Seed Fu Alternative | seedbank gem    | Seed Fu Gem     | Validation-data-as-code | "One will need to maintain the code in the seeds files after every code change. One needs to populate all nedded fields or write logic to create values." |

PR #410, opened 2020-02-04
Bugfix nil class

sarala:
db/schema.rb changes were not manual changes. Is this normal behaviour?


app/serializers/activity_serializer.rb
```diff
@@ -15,7 +15,9 @@ class ActivitySerializer
 
   attribute "prov:wasAttributedTo" do |object|
     url = Rails.env.production? ? "https://api.datacite.org" : "https://api.test.datacite.org"
-    object.username.include?(".") ? url + "/clients/" + object.username : url + "/providers/" + object.username
+    if object.username.present?
```
mfenner: this code doesn't handle object.username.nil? You could do that at the top of the method, e.g. 

```
attribute "prov:wasAttributedTo" do |object|
  return nil if object.username.blank?
```

The rest of the code then doesn't need to change

.gitignore
```diff
@@ -11,6 +11,9 @@
 /db/*.sqlite3
 /db/*.sqlite3-journal
 
+# Ignore seed datafile
```
mfenner: Please don't include this as it is unrelated.

db/schema.rb
```diff
@@ -12,7 +12,7 @@
```
kjgarza: Maybe just don't commit these schema changes. This probably was caused by your local mysql. just do not include them in the commit

sarala:
@mfenner  there are no tests that I can extend. It looks like there is more work to be done around testing for this code. Manuel test is probably easiest for now.

@richardhallett is it okay to merge with codeclimate complaining. I'm not sure how to fix that.

Issue #367, opened 2019-11-29
Bug filter usage events before the date of publication.


kjgarza:
if a repository sends usage events before the date of publication of a dataset, they will be accessible via the API.

To address that we need to filter any usage event before the date of publication.

TODO:
- [ ] transform casing of events in the DB 
- [ ] consider the use of ActiveModelSerializer parser

Related issues:
- https://github.com/datacite/lupo/issues/360
- https://github.com/datacite/lupo/pull/365
- https://github.com/datacite/lupo/issues/352

PR #454, opened 2020-03-18
Attributes `sub` and `obj` of the `event` index are being currently saved as kebab_case thus cannot be accessed

kjgarza:
## Description

Attributes in nested json objects in the events index cannot access because of the casing in bot the Database and the ES. 
This affected the Attributes `sub` and `obj` of the `event` index which are being currently saved as kebab_case. 

See related issues for more detail:
resolves: https://github.com/datacite/lupo/issues/367
resolves: https://github.com/datacite/lupo/pull/365

This PR
A) modifies all the values in the database to camelCase (re-importing all events or transforming all of them) with a Rake task
B) changes the indexJob so the transformation happens there during every index.

## Status
- [x] Ready for Review

## Types of changes

## Non Functional Requirement
- [x] Tests Cover Changes
- [ ] Documentation


kjgarza:
@mfenner i further clarify the title of the PR to make it clearer

app/models/event.rb
```diff
@@ -504,6 +504,39 @@ def self.subj_id_check(options = {})
...
+    response = Event.query(nil,  page: { size: 1, cursor: [] })
+    Rails.logger.warn "[modify_nested_objects] #{response.results.total} events for source datacite-crossref."
+
+    # walk through results using cursor
+    if response.results.total.positive?
+      while response.results.results.length.positive?
+        response = Event.query(nil, page: { size: size, cursor: cursor })
+        break unless response.results.results.length.positive?
+
+        Rails.logger.warn "[modify_nested_objects] modify_nested_objects #{response.results.results.length}  events starting with _id #{response.results.to_a.first[:_id]}."
```
mfenner, line 520: Does this have to be warn or is info also ok?

app/models/event.rb
```diff
@@ -504,6 +504,39 @@ def self.subj_id_check(options = {})
...
+        cursor = response.results.to_a.last[:sort]
+        Rails.logger.warn "[modify_nested_objects] Cursor: #{cursor} "
+
+        ids = response.results.results.map(&:uuid).uniq
```
mfenner, line 524: Why are you not looping through the ids and start a job for each one of them instead of having an extra job for the list of ids?

app/models/event.rb
```diff
@@ -504,6 +504,39 @@ def self.subj_id_check(options = {})
...
+        CamelcaseNestedObjectsJob.perform_later(ids, options)
+      end
+    end
+  end
+
+
+  def self.camelcase_nested_objects(uuid)
+      event = Event.find_by(uuid: uuid)
+      if event.present?
+        subj = event.subj.transform_keys { |key| key.to_s.underscore.camelcase(:lower) }
```
mfenner, line 534: How does this work for nested keys in `subj` and `obj`, or do we not care?

kjgarza: i am not sure if I follow you question; this method transforms  the nested keys in `subj` and `obj` to `camelcase`.  

I have added a test in case that clarifies further how it works

mfenner:
@kjgarza what is the status of this pull request? Are there open questions, or do you want to go ahead and merge?

app/models/event.rb
kjgarza: @mfenner this is the only pending question. any further comments?

mfenner: Thanks. Which test clarifies how it works?

kjgarza: https://github.com/datacite/lupo/pull/454/commits/b6679341975c23c4c53c458442040f4996bf5e3f

kjgarza: and https://github.com/datacite/lupo/pull/454/commits/7d6102cdfb7876b9645140c3ec9ee4f1fa67c076

mfenner: Thanks. Then I think we use different language, as `subj` and `obj` in the tests are not nested objects, but objects without nesting. My whole point was that `transform_keys` does not work on nested objects.

Please go ahead and merge the pull request.

PR #457, opened 2020-03-27
rejects blacklisted JWT

kjgarza:
enables blacklisting specific JWT

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -73,7 +73,7 @@ class ApplicationController < ActionController::API
   def authenticate_user!
     type, credentials = type_and_credentials_from_request_headers
     return false if credentials.blank?
-    raise JWT::VerificationError if ENV["JWT_BLACKLISTED"].split(",").include?(credentials)
+    raise JWT::VerificationError if (ENV["JWT_BLACKLISTED"] || "").split(",").include?(credentials)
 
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -59,6 +59,8 @@
 ENV['MG_DOMAIN'] ||= "mg.datacite.org"
 ENV['HANDLES_MINTED'] ||= "10132"
 ENV['REALM'] ||= ENV['API_URL']
+ENV['JWT_BLACKLISTED'] ||= ENV['JWT_BLACKLISTED']
```
mfenner: This is a fallback to a default value. So this line does not make sense.

app/controllers/application_controller.rb
```diff
@@ -74,6 +73,7 @@ def authenticate_user_with_basic_auth!
   def authenticate_user!
     type, credentials = type_and_credentials_from_request_headers
     return false if credentials.blank?
+    raise JWT::VerificationError if (ENV['JWT_BLACKLISTED'] == credentials)
```
mfenner: I would make  `JWT_BLACKLISTED` a comma-separated string that is converted to a list. There is no reason to assume there will only be one blacklisted JWT.

kjgarza: i was hoping it would be a one-off, but you are right. I'll change that

PR #483, opened 2020-04-22
Improve count performance

sarala:
I've taken the link checking stuff out from the API. @richardhallett can you please check whether it is fine.


sarala:
It would be good to find out which fields are actually used. Is there any other way than breaking changes?

mfenner:
I am not sure how we can find out which facets are used by our users, as the queries will be the same. The only way I see is asking for feedback before or after the change.

PR #489, opened 2020-04-24
Feature rake tasks for data cleaning in our index

kjgarza:
## Description

Two rake tasks to clear erroneous data in our index/db.

- Rake task to remove events with double crossref errors, that have been already labeled. This task can be used once Crossref green-flag us from removing the events. related: #351
- Rake task to schema_version for old dois missing schema version. related: datacite/bracco#225

- rake task
- model method
- background job

## Status
- [x] Ready for Review

## Types of changes

## Non Functional Requirement
- [x] Tests Cover Changes
- [ ] Documentation

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -1776,7 +1776,7 @@ class Doi < ActiveRecord::Base
     query_hsh = {page: { size: 1, cursor: [] } }.merge(filter)
 
     response = Doi.query(query, query_hsh)
-    Rails.logger.warn "#{label} #{response.results.total} Dois with #{label}."
+    Rails.logger.info "#{label} #{response.results.total} Dois with #{label}."
 
     # walk through results using cursor
     if response.results.total.positive?
@@ -1784,12 +1784,14 @@ class Doi < ActiveRecord::Base
         response = Doi.query(query, query_hsh.merge!(page: { size: size, cursor: cursor }))
         break unless response.results.results.length.positive?
 
-        Rails.logger.warn "#{label} #{response.results.results.length}  Dois starting with _id #{response.results.to_a.first[:_id]}."
+        Rails.logger.info "#{label} #{response.results.results.length}  Dois starting with _id #{response.results.to_a.first[:_id]}."
         cursor = response.results.to_a.last[:sort]
-        Rails.logger.warn "#{label} Cursor: #{cursor} "
+        Rails.logger.info "#{label} Cursor: #{cursor} "
 
-        ids = response.results.results.map(&:uid).uniq
-        Object.const_get(job_name).perform_later(ids, filter)
+        ids = response.results.results.map(&:uid)
+        ids.each do |id|
+          Object.const_get(job_name).perform_later(id, filter)
+        end
       end
     end
   end
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -1765,6 +1765,12 @@ class Doi < ActiveRecord::Base
     response.results.total
   end
 
+  # Transverses the index in batches and using the cursor pagination and executes a Job that matches the query and filer
+  # Options:
+  # +filter+:: paramaters to filter the index
+  # +label+:: String to output in the logs printout
+  # +query+:: ES query to filter the index
+  # +job_name+:: Acive Job class name of the Job that would be executed on every matched results 
   def self.loop_through_dois(options)
     size = (options[:size] || 1000).to_i
     cursor = [options[:from_id], options[:until_id]]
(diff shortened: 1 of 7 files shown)


app/models/doi.rb
```diff
@@ -1765,6 +1765,36 @@ def self.transfer(options={})
...
+
+    query_hsh = {page: { size: 1, cursor: [] } }.merge(filter)
+
+    response = Doi.query(query, query_hsh)
+    Rails.logger.warn "#{label} #{response.results.total} Dois with #{label}."
+
+    # walk through results using cursor
+    if response.results.total.positive?
+      while response.results.results.length.positive?
+        response = Doi.query(query, query_hsh.merge!(page: { size: size, cursor: cursor }))
```
mfenner, line 1784: Can this be done simpler, avoiding the double merge?

app/models/doi.rb
```diff
@@ -1765,6 +1765,36 @@ def self.transfer(options={})
...
+        break unless response.results.results.length.positive?
+
+        Rails.logger.warn "#{label} #{response.results.results.length}  Dois starting with _id #{response.results.to_a.first[:_id]}."
+        cursor = response.results.to_a.last[:sort]
+        Rails.logger.warn "#{label} Cursor: #{cursor} "
+
+        ids = response.results.results.map(&:uid).uniq
```
mfenner, line 1791: Why do you use `uniq`, isn't that guaranteed in our index?

app/models/doi.rb
```diff
@@ -1765,6 +1765,36 @@ def self.transfer(options={})
...
+
+    # walk through results using cursor
+    if response.results.total.positive?
+      while response.results.results.length.positive?
+        response = Doi.query(query, query_hsh.merge!(page: { size: size, cursor: cursor }))
+        break unless response.results.results.length.positive?
+
+        Rails.logger.warn "#{label} #{response.results.results.length}  Dois starting with _id #{response.results.to_a.first[:_id]}."
+        cursor = response.results.to_a.last[:sort]
+        Rails.logger.warn "#{label} Cursor: #{cursor} "
```
mfenner, line 1789: Please limit `logger.warn` messages to a minimum, as they are sent to DataDog. Maybe `logger.info` for situations where we need logging.

app/models/doi.rb
mfenner, line 1791: Why do you use `ids` and not `id`, looping through the list of DOIs here instead of the job.

app/models/doi.rb
```diff
@@ -1765,6 +1765,36 @@ def self.transfer(options={})
...
+
+        ids = response.results.results.map(&:uid).uniq
+        Object.const_get(job_name).perform_later(ids, filter)
```
mfenner, line 1792: Isn't the job name fixed for this method? Or do you want to use it as a generic function?

app/models/event.rb
```diff
@@ -511,6 +511,35 @@ def self.label_state_event(event)
...
+    if response.results.total.positive?
+      while response.results.results.length.positive?
+        response = Event.query(query, query_hsh.merge!(page: { size: size, cursor: cursor }))
+        break unless response.results.results.length.positive?
+
+        Rails.logger.warn "#{label} #{response.results.results.length}  events starting with _id #{response.results.to_a.first[:_id]}."
+        cursor = response.results.to_a.last[:sort]
+        Rails.logger.warn "#{label} Cursor: #{cursor} "
+
+        ids = response.results.results.map(&:uuid).uniq
```
mfenner, line 537: See comments ids vs. id above.

app/models/event.rb
```diff
@@ -511,6 +511,35 @@ def self.label_state_event(event)
...
+        Object.const_get(job_name).perform_later(ids, filter)
```
mfenner, line 538: See comment regarding `Object.const_get` above.

lib/tasks/doi.rake
```diff
@@ -89,6 +89,18 @@ namespace :doi do
     Doi.set_minted
   end
 
+  desc "Set schema version"
+  task set_schema_version: :environment do
+    options = {
+      from_id: (ENV["FROM_ID"] || Doi.minimum(:id)).to_i,
+      until_id: (ENV["UNTIL_ID"] || Doi.maximum(:id)).to_i,
+      query: "+aasm_state:findable -schema_version:*",
```
mfenner, line 97: Why `findable`? `registered` also requires valid metadata, and can quickly be turned back to `findable`.

lib/tasks/event.rake
```diff
@@ -79,15 +79,28 @@ namespace :crossref do
 end
 
 namespace :subj_id_check do
-  desc 'checks that events subject node congruency'
-  task :check => :environment do
-    from_id = (ENV['FROM_ID'] || Event.minimum(:id)).to_i
-    until_id = (ENV['UNTIL_ID'] || Event.maximum(:id)).to_i
-    
+  desc "checks that events subject node congruency"
```
mfenner, line 82: Can you use a description that is easier to understand?

app/models/doi.rb
kjgarza: we need this to do the function generic.

app/models/event.rb
kjgarza: we need this to do the function generic.

app/models/doi.rb
kjgarza: we can do that

lib/tasks/event.rake
kjgarza: this tasks is for another purpose, i was just fixing linting.

nevertheless i added a better description and moved into another task to get a better context of it use

PR #494, opened 2020-04-30
Graphql autoload fixes

mfenner:
This pull request fixes the issues with auto-loading GraphQL that we are seeing since Sunday. It is essentially reverting back to what we had working until Sunday, using a "hack" to make auto-loading GraphQL files work. The official approach to setup GraphQL that the changes on Suday tried to implement, don't work in a production environment, despite passing 290 tests in the test environment.

Auto-loading issues are a long-running issue with graphql-ruby, see for example https://github.com/rmosolgo/graphql-ruby/issues/1245. We will continue to use the "hack" of eager loading the required files in `config/application.rb`:

```
config.paths.add Rails.root.join('app', 'graphql', 'types').to_s, eager_load: true
config.paths.add Rails.root.join('app', 'graphql', 'mutations').to_s, eager_load: true
```

We can revisit after upgrading to Rails 6, which uses a new `zeitwerk` library for auto-loading.


app/graphql/types/model_type.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class Types::ModelType < Types::BaseObject
-  implements Types::DoiItem
+class ModelType < BaseObject
+  implements DoiItem
```
kjgarza, line 4: I think the previous version was  ` implements ::DoiItem` but i'm unsure if that matter much

mfenner: That was used inconsistently in the previous version.

Gemfile
```diff
@@ -58,12 +58,12 @@ gem "rack-utf8_sanitizer", "~> 1.6"
 gem "turnout", "~> 2.5"
 gem "audited", "~> 4.8"
 gem "git", "~> 1.5"
-gem "graphql", "~> 1.9.16"
```
kjgarza, line 61: There are many gems updates, if we are reversing maybe it might be worth keeping the old ones. I remember there was in the pass a problem related to graphql gem update that affected the api and we couldn't see it.

app/graphql/types/model_type.rb
kjgarza, line 4: I think the previous version was `implements ::DoiItem` but i'm unsure if that matter much

Gemfile
mfenner: The specific update is to the latest version of the `graphql` gem. This is now finally possible, as the `apollo-federation` gem now supports `graphql` version 1.10.

kjgarza:
locally seems to work 
image

config/application.rb
```diff
@@ -63,10 +63,11 @@
 module Lupo
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.2
+    # config.load_defaults 5.2
```
richardhallett, line 66: Do we no longer need the default configs?

I believe these are the defaults that get loaded:

3.17.2 For '5.2', new defaults from previous versions below and:

    config.active_record.cache_versioning: true
    config.action_dispatch.use_authenticated_cookie_encryption: true
    config.active_support.use_authenticated_message_encryption: true
    config.active_support.use_sha1_digests: true
    config.action_controller.default_protect_from_forgery: true
    config.action_view.form_with_generates_ids: true

3.17.3 For '5.1', new defaults from previous versions below and:

    config.assets.unknown_asset_fallback: false
    config.action_view.form_with_generates_remote_forms: true

3.17.4 For '5.0':

    config.action_controller.per_form_csrf_tokens: true
    config.action_controller.forgery_protection_origin_check: true
    ActiveSupport.to_time_preserves_timezone: true
    config.active_record.belongs_to_required_by_default: true
    config.ssl_options: { hsts: { subdomains: true } }

mfenner: I will re-enable the 5.2 default configs, although that probably doesn't make a difference.

PR #496, opened 2020-05-03
Switch to puma application server

mfenner:
This pull request replaces `passenger` with `puma` as the Ruby application server. This is a big change and needs more work and testing. The reasons for this change are multiple:

* Puma is faster
We are working on improving the performance of the API and should not forget the application server. 

* Puma makes it easier to generate Cloudwatch metrics
I have spent a full day last week getting passenger (via `passenger-config` and `passenger-status`) to integrate with AWS Cloudwatch Metrics. It is doable, but there are major issues to resolve.

Cloudwatch metrics can be used for auto-scaling and will give a much better indicator of API status compared to CPU or memory usage.

* We can start to use generic Docker images
This pull request uses the very similar Docker image `phusion/baseimage`, but going forward we have more flexibility, including using Alpine Linux with its much smaller footprint.


config/puma.rb
```diff
@@ -0,0 +1,26 @@
...
+# Set up socket location
+bind "unix://#{shared_dir}/sockets/puma.sock"
+
+# Set master PID and state locations
+pidfile "#{shared_dir}/pids/puma.pid"
+state_path "#{shared_dir}/pids/puma.state"
+activate_control_app
+
+on_worker_boot do  
+  require "active_record"
```
kjgarza, line 23: locally i get a 504 error  

```
/var/www/vendor/bundle/ruby/2.5.0/gems/aws-sdk-s3-1.63.0/lib/aws-sdk-s3/bucket.rb:823:in `extract_name': Cannot load `Rails.config.active_storage.service`: (ArgumentError)
web_1            | missing required option :name

```

Dockerfile
```diff
@@ -1,64 +1,56 @@
-FROM phusion/passenger-full:1.0.9
-LABEL maintainer="kgarza@datacite.org"
+FROM phusion/baseimage:0.11
+LABEL maintainer="mfenner@datacite.org"
+
```
kjgarza, line 3: "`Live-reload`" doesn't seem to work anymore. When running docker-compose, If I make changes in the code; they do not seem reflected in the running app. One needs to rebuild the whole container.

richardhallett:
An initial glance the difference between Puma and Passenger as a choice seems to be on following points:

### Performance 
Single threaded multi process (passenger community) vs single process multi threaded (puma)

This is a conceptually different way to do concurrency and in-app multithreading can come with a variety of interesting problems as you're sharing memory, it can also be faster, multi proccess model gives you a bit easier management time as it's all self contained. 

I think for our use case it's debatable if Puma would be faster, I'm not convinced that our bottlenecks i.e. long running API requests rather than throughput would be definitely faster under a multithreaded model, though it is a possibility, the decreased memory usage might be nice to have as well on docker containers, it would need further investigation.

### Error handling
There is also the difference in how they do fault tolerance, which while it might not be an issue for us giving we have other checks in place, there is the possibility this could change how the services behave.

### Cloudwatch metrics
On this subject, I can't speak for the difficulties in getting working, nor of it's advantages, so if it is crucial we need it, then switching to something better supported would make sense. However I do note that we already have a lot of metrics available through the datadog integrations.

### Docker image consistency
What would we do about the other servers we have running, change them as well? Or do we accept that we have different, which is potentially okay, just something to consider.
I note that for Viringo (OAI-PMH) I explicitly went with passenger not because it was nescessarily the best choice for a python web server but because it was what we've used everywhere else in our stack.

I do however support in principle using smaller docker base images, there is sometimes a little too much bloat in the passenger images for our use cases.

## General thoughts
My concern would be this is a large change that could have unknown side effects due to the model change, it is definitely worth considering these changes as they could have a significant performance or even cost saving (if the images can be smaller), but it might be worth spending some extra effort planning how we can test this.

### Alternatives:
If we're looking at replacing the Ruby application server we might also want to consider Unicorn behind Nginx this would give us a lower footprint as well.

mfenner:
@richardhallett thank you for the detailed feedback. Let's discuss more in our dev meeting today.

mfenner:
After an extensive team discussion we decided to:

* stick with passenger for now, until we have clear indicators for performance bottlenecks. For this we will
* send passenger metrics to Cloudwatch
* We are happy to use other Docker base images going forward, but not at this point in time.

Closing the pull request.

PR #497, opened 2020-05-03
proper support for alternate identifiers

mfenner:
The `identifiers` object for DOIs is a combination of the DOI expressed as URL, and any alternate identifiers. Unfortunately it is hard to update this combination of metadata. We are therefore providing proper support for alternate identifiers, while avoiding breaking changes.

This pull request provides backend support for https://github.com/datacite/bracco/pull/407.


app/serializers/doi_serializer.rb
```diff
@@ -6,7 +6,7 @@ class DoiSerializer
...
+  attributes :doi, :prefix, :suffix, :identifiers, :alternate_identifiers, :creators, :titles, :publisher, :container, :publication_year, :subjects, :contributors, :dates, :language, :types, :related_identifiers, :sizes, :formats, :version, :rights_list, :descriptions, :geo_locations, :funding_references, :xml, :url, :content_url, :metadata_version, :schema_version, :source, :is_active, :state, :reason, :landing_page, :view_count, :views_over_time, :download_count, :downloads_over_time, :reference_count, :citation_count, :citations_over_time, :part_count, :part_of_count, :version_count, :version_of_count, :created, :registered, :published, :updated
```
kjgarza, line 9: I think, as it is, the serializer will show both identifier and alternateIdentifier. I am not sure that is necesary.

app/controllers/dois_controller.rb
```diff
@@ -689,14 +691,23 @@ def safe_params
     read_attrs_keys.each do |attr|
       p.merge!(attr.to_s.underscore => p[attr] || meta[attr.to_s.underscore] || p[attr]) if p.has_key?(attr) || meta.has_key?(attr.to_s.underscore)
     end
+
+    # handle alternate_identifiers
+    p[:identifiers] = Array.wrap(p[:alternate_identifiers]).map do |a|
```
kjgarza, line 696: I think this will replace `identifier` with `alternateIdentifier`. Shouldn't we cope with both cases?

app/controllers/dois_controller.rb
```diff
@@ -689,14 +691,23 @@ def safe_params
...
+      { "identifierType" => a["alternateIdentifierType"], "identifier" => a["alternateIdentifier"] }
```
kjgarza, line 697: Doesn't **bolognese** also need to be adjusted for this change ?? https://github.com/datacite/bolognese/blob/faddefb6edebdcd4060fd75e67c26176e13f5706/lib/bolognese/datacite_utils.rb#L124

spec/requests/dois_spec.rb
```diff
@@ -186,6 +186,8 @@
```
kjgarza, line 186: Integration Tests  (request/doi_spec.rb) need to be adjusted to accommodate the changes.

mfenner: Please see the tests in the pull request.

app/controllers/dois_controller.rb
mfenner: Please explain.

app/controllers/dois_controller.rb
mfenner: What is the other case? Users submitting `identifiers` to the REST API?

app/serializers/doi_serializer.rb
mfenner: This is to avoid breaking changes to the API. It is of course duplication.

mfenner:
@kjgarza can you specify the breaking change in the API? 

And how do we best accomodate having a stable REST API with a fix we need to make to release the DOI form to production?

spec/requests/dois_spec.rb
kjgarza: I mentioned because Travis is failing on `rspec ./spec/requests/dois_spec.rb:181 # dois GET /dois/:id when the record exists returns the Doi`

app/controllers/dois_controller.rb
kjgarza: exactly. This was previously mentioned in support https://app.frontapp.com/open/msg_bban6tf

mfenner: To be clear, the support message is not about submitting `identifiers`. We can clarify that `alternateIdentifiers` should be used.

app/controllers/dois_controller.rb
kjgarza: well this is mostly a question. Since bolognse makes metadata transformation and used in various places. I was wondering whether this change will impact it.

kjgarza:
well since the only change that user could notice would be in the doi parameters in the controller, we could condition that part:

```
    # handle alternate_identifiers
    p[:identifiers] = Array.wrap(p[:alternate_identifiers]).map do |a|
      { "identifierType" => a["alternateIdentifierType"], "identifier" => a["alternateIdentifier"] }
    end.compact
```

We know the `source` by then. One can add condition that when Fabrica is sending metadata it consider the new code but when a client does it (any other source) then uses the old approach.

mfenner:
I feel that this adds complexity to an already complex metadata registration. Plus, these changes also help with metadata registration for other users. I think going forward we want to phase out `identifiers`.

richardhallett:
Although I'm late to this PR, I wanted to echo the sentiments by Kristian regarding changing the API I also noticed a few things as below.

### Data loss potential
The main use-case that I think is a little subtle is potential data loss:

I just ran a test and the scenario is this:

1. Existing DOI that has existing "identifiers"
2. GET that DOI
3. Immediately Update the DOI using the response (i.e. as allowed by JSON:API)
4. The data in identifiers is removed because alternateIdentifiers is sent as empty.

### Create/Update doesnt work as before with identifiers
The other usecase is as Kristian mentioned, those users who are just were already expecting to update via "identifiers" attribute and it won't actually do anything, so if they have an existing integration they may not even notice that there metadata is not creating/updating with alternate identifiers.

### DataCite JSON format 
This format returned from the API uses "identifiers" still, is this as expected? I think maybe that is what Kristian meant by changing bolognese to match.

### Potential bug
I think the returned "identifiers" is returning null for the data, I will raise this as a separate issue, but I believe it is the mapping code in safe params using alternate_identifiers not alternateIdentifiers.
https://github.com/datacite/lupo/issues/498

### Final thoughts
If we're making this a minor release, then my preference would be to handle identifiers and alternateIdentifiers as though they are just synonyms of each other, i.e. it does not matter if you send one or the other.

I do appreciate the argument for it already being complex, but I wanted to make sure we're fully aware of this change explicitly.

mfenner:
Thanks for this comment @richardhallett. Let me stress the fundamental issue with our `identifiers` implementation, which I didn't see when implementing this originally: `identifiers` contains the DOI for the resource, and that is a read-only property.

The immediate issue to solve is allowing users to update `alternateIdentifiers` using form or REST API, ideally without any breaking changes. Medium-term (i.e. next API version) it is probably better to get rid of `identifiers`. Short-term we probably will have `identifiers` and `alternateIdentifiers` in parallel.

From what you say there might be a little bit more work to do, mainly in the controller and serializer.

richardhallett:
Ah yes when I left the comment, I was missing that part of the PR, you're right the current implementation is confusing which contains both.

In addition we are using the combination in places like OAI-PMH, though this won't actually break anything at present, though if the DOI is removed, we might need to refactor to use alternateIdentifiers and add the DOI in explicitly (as OAI-PMH DC representation expects all identifiers, not just alternative)

Issue #503, opened 2020-05-07
Improve performance of events aggregations


mfenner:
* review the list of aggregations we want to provide
* limit results to 10

mfenner:
Deleting the following aggregations:

* dois
* states
* pairings

Simplifying the aggregations and limiting results to 10.

mfenner:
Fixed via #505.

PR #505, opened 2020-05-07
Improve performance of events api

mfenner:
This pull request makes changes described in #503, improving the performance of aggregations and facets in the events API endpoint.


app/controllers/events_controller.rb
```diff
@@ -139,16 +139,12 @@ def index
 
       render json: EventSerializer.new(results, options).serialized_json, status: :ok
     else
-      aggregations = params.fetch(:aggregations, "") || ""
```
kjgarza, line 142: is it truly necessary to stop supporting the `aggregation selection` functionality? This functionality doesn't have an impact in performance.

app/controllers/events_controller.rb
```diff
@@ -139,16 +139,12 @@ def index
...
-
-      sources = total.positive? && aggregations.blank? || aggregations.include?("query_aggregations") ? facet_by_source(response.response.aggregations.sources.buckets) : nil
-      prefixes = total.positive? && aggregations.blank? || aggregations.include?("query_aggregations") ? facet_by_source(response.response.aggregations.prefixes.buckets) : nil
-      citation_types = total.positive? && aggregations.blank? || aggregations.include?("query_aggregations") ? facet_by_citation_type(response.response.aggregations.citation_types.buckets) : nil
-      relation_types = total.positive? && aggregations.blank? || aggregations.include?("query_aggregations") ? facet_by_relation_type(response.response.aggregations.relation_types.buckets) : nil
-      registrants = total.positive? && aggregations.blank? || aggregations.include?("query_aggregations")  ? facet_by_registrants(response.response.aggregations.registrants.buckets) : nil
-      pairings = total.positive? && aggregations.blank? || aggregations.include?("query_aggregations") ? facet_by_pairings(response.response.aggregations.pairings.buckets) : nil
```
kjgarza, line 149: I think the main performance improvement on this PR is about removing `states` and `pairings` from the aggregations. And that can be easy achieved by just removing them out of the "query_aggregations" hash.

app/controllers/events_controller.rb
mfenner: The overall code for Elasticsearch aggregations is overly complex and I am not sure what value it adds. Right now the top priority is performance.

app/controllers/events_controller.rb
mfenner: The performance of the events API is very poor, and we will not get away with minor changes. This is just the first step.

sarala:
I don't understand all of it but I can see bucket sizes being reduced etc and works locally.

PR #511, opened 2020-05-12
Implement GraphQL Relay pagination

mfenner:
## Purpose
This pull request implements proper pagination in GraphQL. The `graphql` gem added proper support for [relay pagination](https://relay.dev/graphql/connections.htm) in version `1.10`, and we are finally using this version - since the `apollo-federation` gem started to support this version in [April](https://rubygems.org/gems/apollo-federation/versions/1.0.4).

## Approach
The main change is support for custom paginators and this pull request adds paginators for Elasticsearch and REST APIs (we have a GraphQL wrapper for the ORCID, re3data, ROR and Crossref Funder ID APIs). Relay expects a cursor-based pagination, and this aligns well with our Elasticsearch cursor-based pagination.

- [ ] There are still issues and broken tests with properly passing along pagination parameters for REST API pagination.
- [ ] do we need bidirectional pagination, as supported by Relay, but not yet our Elasticsearch implementation?
- [ ] do we need to support sorting? Sorting is fixed in our cursor-based pagination
- [ ] do we need to support pagination by page numbers? Do we need to show links to the next 10 pages?
- [ ] there is currently a lot of code duplication for GraphQL connections. I will tackle this after the basic implementation is working correctly.

## Learning
This is another major step towards making the GraphQL API a full replacement for our REST API.


richardhallett:
An opinion on the pre merge questions:

1. Bi-directional pagination - Given that ES doesn't support it, I think we can probably skip it for now, as it might be tricky, additionally for most part I think if you're looking at data after you've narrowed down your filtering, it's unlikely you need to step back.

2. Sorting - As like the previous, depending on scenario, given for this is aimed more at querying and examining results, you're more likely to sort your work post graphql results. But I could be wrong.

3. See above comment, but in general pagination back and forth and numbering is more useful in a scenario where you're going through literal pages, e.g. a blog, or a product listing.

mfenner:
@richardhallett Elasticsearch supports `search_before`, see https://github.com/elastic/elasticsearch/issues/29449

mfenner:
I have made significant changes to the code to properly support pagination with Elasticsearch and calls to external REST APIs. There are still some minor issues regarding `hasNextPage`, but that should not stop us from releasing this

mfenner:
I have implemented Relay pagination without bidirectional linking, so only forward-linking using `after` and `endCursor`, no `before` or `startCursor`.

spec/models/doi_spec.rb
```diff
@@ -523,22 +523,23 @@
...
-      Doi.import
-      sleep 2
-    end
-
-    it "transfer all dois" do
-      response = Doi.transfer(client_id: client.symbol.downcase, target_id: target.symbol.downcase, size: 3)
-      expect(response).to eq(5)
-    end
-  end
+  # TODO issue with search_after
```
richardhallett, line 526: Does this mean that transfers wont work with these changes?

mfenner: My guess is that this is unrelated. We need to test further.

PR #512, opened 2020-05-12
new stats query

sarala:
I have written a simplified query for collecting stats for doi but I'm not sure whether it is correct. I've just done it for 'admin' 'doi_count'.

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -235,7 +235,7 @@ class ProvidersController < ApplicationController
       providers = provider_count(consortium_id: nil)
       clients = client_count(provider_id: nil)
       dois = doi_count(provider_id: nil)
-      resource_types = resource_type_count(provider_id: nil)
+      # resource_types = resource_type_count(provider_id: nil)
       # citations = nil # citation_count(provider_id: nil)
       # views = nil # view_count(provider_id: nil)
       # downloads = nil # download_count(provider_id: nil)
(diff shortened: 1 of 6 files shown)


app/models/doi.rb
```diff
@@ -675,6 +675,24 @@ def self.find_by_id(id)
...
+  def self.stats_query(query, options={})
+    aggregations = {created: { date_histogram: { field: 'created', interval: 'year', format: 'year', order: { _key: "desc" }, min_doc_count: 1 },
+                    aggs: { bucket_truncate: { bucket_sort: { size: 10 } } } },
+                   }
+
+    from = 0
+    sort = [{ created: "asc", uid: "asc" }]
+
+    __elasticsearch__.search({
+      size: options.dig(:page, :size),
```
mfenner, line 687: You don't need from and sort, so I would leave them out. I would hardwire `size: 0`, as you are only interested in the aggregations.

I would write `query` so that the function also supports stats for a single provider or client.

The aggregation should have a `size` larger than 10, so that we include all DataCite DOIs. We currently start at 2010 (when DataCite was founded), so 11 years.

Have you looked at `track_total_hits`? Not sure what the best advice is for your query. https://www.elastic.co/guide/en/elasticsearch/reference/master/search-request-body.html#request-body-search-track-total-hits

app/controllers/providers_controller.rb
```diff
@@ -235,23 +235,23 @@ def stats
       providers = provider_count(consortium_id: nil)
       clients = client_count(provider_id: nil)
       dois = doi_count(provider_id: nil)
-      resource_types = resource_type_count(provider_id: nil)
+      # resource_types = resource_type_count(provider_id: nil)
```
mfenner, line 238: In the next step, this could be integrated with doi_count into a single elasticsearch query and two aggregations.

mfenner:
Please fix the one failing test (failing because a method is commented out in this pull request).

app/models/doi.rb
sarala: @mfenner, could you please have a look at the query that I'm constructing. It is not quite working but I haven't been able to figure out why yet.

PR #513, opened 2020-05-13
implement graphql actor interface

mfenner:
## Purpose
This pull request implements an `actor` entity that can either be a person, research organization or funder. This allows lookup by identifier or queries across all three services. A query for `Cambridge University` will for example find research organizations and funders with that name, but also people with that affiliation in their ORCID record.

## Approach
This is implemented as GraphQL `interface. There are four common fields: id, type, name, alternateName. Additional fields can be returned if they exist for that particular entity, e.g. 

```
{
  actors(query: "Cambridge University") {
      totalCount
      nodes {
        id
        type
        name
        alternateName
        ... on Person {
          affiliation {
            name
          }
        }
      }
    }
```

- [ ] consider implementing a `pid` type that accepts all PIDs used in the GraphQL API

## Learning
The main architecture to make was whether actors should be implemented as `interface` or `union`. Given that there are a few common fields, `interface` was the better choice and for example makes query construction easier.

* https://github.com/rmosolgo/graphql-ruby/blob/master/guides/type_definitions/interfaces.md
* https://artsy.github.io/blog/2019/01/14/graphql-union-vs-interface/

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -3,7 +3,7 @@
 class QueryType < BaseObject
   extend_type
 
-  field :members, MemberConnectionType, null: false, connection: true, max_page_size: 1000 do
+  field :members, MemberConnectionType, null: false, connection: true do
     argument :query, String, required: false
     argument :first, Int, required: false, default_value: 25
   end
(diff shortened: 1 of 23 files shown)


app/graphql/types/query_type.rb
```diff
@@ -3,7 +3,7 @@
 class QueryType < BaseObject
   extend_type
 
-  field :members, MemberConnectionType, null: false, connection: true, max_page_size: 1000 do
```
kjgarza, line 6: In general it looks ok. However I do think it would help as a reviewer that there different commits for different things. I guess this is about pagination and it has nothing to do with the actor interface. its totally fine its in the same PR but maybe use more than one commit so its easier to read the code during the review

app/models/person.rb
```diff
@@ -70,21 +71,8 @@ def self.parse_message(message: nil)
       orcid: orcid,
       given_name: given_name,
       family_name: family_name,
-      other_names: other_names,
```
kjgarza, line 73: Does this change would `break` the public interface for someone that has integrated with `other_names` already? if it does then i would think this will need to be documented or managed so other integrations do not `fail`. That probably will dictate wether this change is a 'MINOR' or a 'MAJOR' release

mfenner:
Thank you for your feedback @richardhallett and @kjgarza.

GraphQL APIs typically don't do versioning, as you can avoid breaking changes by just adding functionality. `other_names` for `person` is an edge case, as that API was released 7 days ago.

PR #493, opened 2020-04-29
Feature  Repository Transfer

kjgarza:
## Description

Implements API functionality to transfers Repositories/clients to another provider.

payload for transfer
```
{
  "data": {
    "type": "clients", 
    "attributes": {
      "mode": "transfer", 
      "targetId": "TARGET.REPOSITORY"
    }
  }
}
```

resolves: https://github.com/datacite/lupo/issues/486

## List of General Components affected
- Client and doi models
- new background jobs
- small fix in seeding task
- abilities and authentication

## Status
- [x] Ready for Review

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Non Functional Requirement
- [x] Follows the code style of this project.
- [x] Tests Cover Changes
- [x] All new and existing tests passed.
- [ ] Documentation

PR #514, opened 2020-05-14
Fix prefixes build order in repository transfer

kjgarza:
## Purpose

Repository transfer was not assigning prefixes to the repository at the end of the transfer. Test were not failing because the order of building prefixes associations was not correct.

related: https://github.com/datacite/lupo/pull/493

## Approach

- changed order of data building in tests
- add recreate association step in transfer prefix method

## Learning

ProviderPrefixes dependant association remove ClientPrefixes automatically.
https://github.com/datacite/lupo/blob/a59d225/app/models/provider_prefix.rb#L12

ProviderPrefixes building order when creating factory data matters.

## Status
- [x] Ready for Review

## Types of changes

### DIFF
--- a/spec/models/client_spec.rb
+++ b/spec/models/client_spec.rb
@@ -27,9 +27,11 @@ describe Client, type: :model do
     let!(:prefixes) { create_list(:prefix, 3) }
     let!(:prefix) { prefixes.first }
 
-    let!(:client_prefix)  { create(:client_prefix, client: client, prefix: prefix) }
+    ### Order is important in creating prefixes relations
     let!(:provider_prefix)  { create(:provider_prefix, provider: provider, prefix: prefix) }
     let!(:provider_prefix_more) { create(:provider_prefix, provider: provider, prefix: prefixes.last) }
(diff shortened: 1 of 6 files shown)


spec/models/client_spec.rb
```diff
@@ -101,9 +105,10 @@
   describe "Client prefixes transfer" do
     let!(:prefixes) { create_list(:prefix, 3) }
     let!(:prefix) { prefixes.first }
-    let!(:client_prefix)  { create(:client_prefix, client: client, prefix: prefix) }
-    let!(:provider_prefix)  { create(:provider_prefix, provider: provider, prefix: prefix) }
+     ### Order is important in creating prefixes relations
```
richardhallett, line 108: This comment is a little confusing, it's not so much the order is important, but the fact the client_prefix needs the provider_prefix id in order to relate it, if you left it in the original order while trying to reference provider_prefix.uid this would fail to build.

kjgarza: actually it wouldn't fail, that is why i did it caught it the first time. The factory has a 'create' strategy so if there is no `provider_prefix_id` it will create one, but not the one that one wants

PR #515, opened 2020-05-15
fix: add subjects to meta

kjgarza:
## Purpose
blog doesn't show tag anymore because subject were removed from the api meta object
addresses: https://github.com/datacite/blog/issues/219

## Approach

return subjects

## Status
- [x] Ready for Review

## Types of changes


sarala:
@mfenner, yes, @kjgarza and I had a quick discussion over it. I didn't consider getting rid of the subject facets in the blog. If that's possible, I think it's a better solution.

kjgarza:
we will remove the feature as part of performance improvements.

kjgarza:
re-opening as it has more uses

Issue #518, opened 2020-05-17
Normalize database tables to follow Rails conventions


mfenner:
The `lupo` service is using a legacy database schema from the previous MDS service. Since retiring Solr in December 2019 this schema is no longer needed. We should convert the schema to use Rails naming conventions to simplify the `lupo` configuration:

* table names are model names in plural (as model names will be changed in a separate issue, we will go with `members` and `repositories`)
* remove obsolete database columns that are no longer used, e.g. `doi_quota_allowed`
* rename `created` and `updated` datetime columns to `created_at` and `updated_at`, to allow Rails to automatically update those columns
* add `uid` columns for `members` and `repositories` that will replace `symbol` going forward.

Issue #517, opened 2020-05-17
Refactor backend services for common DOI search


mfenner:
We need to make the following changes to our backend services in `datacite/lupo` to properly handle a common DOI search:

* changes to the `Doi` model, so that DOIs from other DOI registration agencies are treated differently than DataCite DOIs, including that there are no repositories or members linked to the DOI
* separate database table.
* separate Elasticsearch index

For the separate database table we will use the horizontal database sharding that will be released with Rails 6.1. We will thus need to do the following steps as well:

* normalize database tables to follow Rails conventions
* align model names, using `member` and `repository`
* upgrade to Rails 6.0
* upgrade to Rails 6.1

As this is significant extra work and Rails 6.1 will not be released until July 2020, the separation of database tables will be completed in a separate epic.

PR #526, opened 2020-05-17
Database normalization

mfenner:
## Purpose
This pull request implements issue **Normalize database tables to follow Rails conventions** described in #518. Briefly, database and field names are aligned with Rails conventions. This is part of the backend work on common DOI search, summarized in #517.

Reviewers are asked to focus on the changes in the database tables, e.g. whether the renaming or removing of specific fields could cause an issue.

## Approach
In addition to renaming tables and fields, and removing obsolete fields, this pull request also makes sure that older migrations run properly (e.g. when setting up a system from scratch), as the [strong migrations gem](https://github.com/ankane/strong_migrations) that we started using recently breaks some of these migrations.

- [ ] Merging will require to immediately run the database migrations. For the production system this requires a planned service outage, which is planned for May 23.

## Learning
The required code changes were easier than anticipated. Good test coverage is essential for this pull request.


mfenner:
@kjgarza thanks for the comment. I think you highlight a common problem, but don't forget that this pull request is about at least being consistent with naming across model, view and controller, which we have never been in this application.

Issue #499, opened 2020-05-06
Missing contacts in CSV export


mfenner:
### Expected Behaviour
All contacts should be exported from Fabrica for the sync to SF.

### Current Behaviour
If there is the same email address in two different organizations in Fabrica, only one contact is exported.

### Steps to Reproduce
As an example, paloma.arraiza@tuwien.ac.at is a contact under TUW (consortium organization) and TUWIENCO (consortium lead).
Although when the contacts export, the csv file only contains one listing of paloma.arraiza@tuwien.ac.at and is listed against the Fabrica ID TUWIEN (not related to any of the above organizations - consortium organization or consortium lead).

### Context (Environment)
Affects the sync with SF.

### Hypothesis
The export seems to exclude duplicates by email address, and should rather use Fabrica contact ID.
The export has an unknown Fabrica ID (e.g. TUWIEN), the reason is unknown or possibly could be a character limit on the Fabrica ID export field (although unlikely).

mfenner:
Fixed by #528.

PR #528, opened 2020-05-18
Allow export of multiple contacts with same email but different provider id

mfenner:
## Purpose
This pull request addresses #499, supporting the CSV export of multiple contacts with the same email address, but different provider id.

## Approach
The contact list was built using the email address as unique key. The pull request implements the fabricaId (providerId + email) as unique key.

- [ ] the outcome is multiple records of the same person in for example Salesforce, but that is intentional.
- [ ] at some point `add_contact` should be refactored into a standalone method, which simplifies unit testing.

## Learning
Only small changes in the code needed.


mfenner:
Closes #499.

PR #532, opened 2020-05-22
prefix ES indexes to support multiple services in one cluster

mfenner:
## Purpose
Support indexes for test/sandbox service in the same Elasticsearch cluster

## Approach
Use an environmental variable to prefix index name

None.


app/models/event.rb
```diff
@@ -106,7 +106,7 @@ class Event < ActiveRecord::Base
   attr_accessor :container_title, :url
 
   # use different index for testing
-  index_name Rails.env.test? ? "events-test" : "events"
+  index_name Rails.env.test? ? "events-test" : "events-#{ENV["ES_PREFIX"]}"
```
kjgarza, line 109: This means that the `events` **production**  index will be now called `events-production_v1` (instead of just `events_v1`), right? Do we need to then recreate all the indexes before this goes into production?

mfenner: Yes, all indexes need to be re-created. We could make this opt-in, i.e. an empty `ES_PREFIX` ENV variable would not change the name of the index.

kjgarza: I don't have a strong preference on the matter. Having said that re-creating all the indexes does seems a time consuming task that needs to be considered for the release.

PR #535, opened 2020-05-27
fix rake task

mfenner:
## Purpose
The rake task to add target_dois to all events is broken.

## Approach
Fix.

None.

## Learning
No easy way to not check multiple times.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -594,7 +594,7 @@ class Event < ActiveRecord::Base
         cursor = response.results.to_a.last[:sort]
         Rails.logger.info "#{label} Cursor: #{cursor} "
 
-        ids = response.results.results.map(&:uuid)
+        ids = response.results.results.map(&:uuid).uniq
         ids.each do |id|
           Object.const_get(job_name).perform_later(id, filter)
         end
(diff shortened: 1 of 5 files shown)


app/models/event.rb
```diff
@@ -379,26 +379,23 @@ def self.update_crossref(options = {})
...
   end
 
-  def self.update_target_doi(options = {})
-    size = (options[:size] || 1000).to_i
-    cursor = (options[:cursor] || [])
-    target_relation_type_id = options[:target_relation_type_id]
-
-    response = Event.query(nil, target_relation_type_id: target_relation_type_id, page: { size: 1, cursor: [] })
-    Rails.logger.info "[Update] #{response.results.total} events with target_relation_type_id #{target_relation_type_id.to_s}."
+  def self.update_target_doi(cursor: nil)
```
kjgarza, line 382: We could keep the code a bit DRY if we use `loop_through_events`  instead of `update_target_doi`.  And we can remove `update_target_doi` all together.

https://github.com/datacite/lupo/blob/b31a661/app/models/event.rb#L596

mfenner: https://github.com/datacite/lupo/blob/b31a661/app/models/event.rb#L598 should use Event instead of Doi, right?

mfenner: @kjgarza Can you push your suggested changes to the branch. That might be easier.

PR #536, opened 2020-05-27
import dois by client rake task

mfenner:
## Purpose
Put all DOIs from a particular client into Elasticsearch. Main use case is DOis not transferred in a DOI transfer as DOis were not indexed in Elasticsearch.

## Approach
Using a rake task. 

- [ ] Potentially more error reporting

## Learning
This does not work so well for clients with many DOIs. Will test performance.

### DIFF
--- a/lib/tasks/client.rake
+++ b/lib/tasks/client.rake
@@ -70,7 +70,7 @@ namespace :client do
   end
 
   desc 'Import DOIs by client'
-  task :import_all_dois => :environment do
+  task :import_dois => :environment do
     if ENV['CLIENT_ID'].nil?
       puts "ENV['CLIENT_ID'] is required."
       exit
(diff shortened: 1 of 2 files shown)


lib/tasks/client.rake
```diff
@@ -83,26 +83,8 @@ namespace :client do
...
-
-    client = Client.where(deleted_at: nil).where(symbol: ENV['CLIENT_ID']).first
-    if client.nil?
-      puts "Client not found for client ID #{ENV['CLIENT_ID']}."
-      exit
-    end
-
-    # import DOIs for client
-    # puts "#{client.dois.length} DOIs will be imported."
-    client.import_missing_dois
```
richardhallett, line 105: Maybe remove the corrosponding client function if this is meant to replace.

PR #542, opened 2020-05-29
Facet for pid entities

mfenner:
## Purpose
Provide a facet in GraphQL for services that shows the PID entities (dataset, publication, funder, instrument, etc.) covered by the service.

## Approach
Using the subject field and the subjectScheme `pidEntity`.

- [ ] correctly filter out all subjects not using subjectScheme `pidEntity`

## Learning
Using filters in Elasticsearch aggregations.


mfenner:
@richardhallett PID entities must be from a controlled list defined in the aggregation. Without this, the aggregation would return all subjects for that DOI.

PR #553, opened 2020-06-03
Disable ActiveJob logging

mfenner:
## Purpose
ActiveJob logging is too verbose at log level INFO.

## Approach
Disable logging. Filtering out the logs doesn't work for `development` and `test`. ActiveJob logging is not essential.


sarala:
This looks fine. I have excluded activejob logging in Datadog log index as well.

mfenner:
@sarala can you please approve if there are no further comments?

PR #555, opened 2020-06-03
Add spdx support to api

mfenner:
## Purpose
Add support for `rights_list` properties added in schema 4.2:

* rightsIdentifier
* rightsIdentifierScheme
* schemeUri

## Approach
Basic implementation in bolognese gem. Requires re-indexing. Adding a `licenses` facet, and a `license` query parameter.


spec/requests/dois_spec.rb
```diff
@@ -3081,9 +3087,19 @@
       end
 
       it 'updates the Doi' do
-        patch "/dois/#{doi.doi}", update_attributes, headers
+        patch "/dois/#{doi.doi}?license=CC0-1.0", update_attributes, headers
```
kjgarza, line 3090: I'm not sure if I don't understand this, Could you help me to understand why the licence need to passed as an attribute in this test ?

mfenner: This is testing that the query parameter for license works.

mfenner: Thanks for catching this. Will fix.

PR #556, opened 2020-06-04
Gracefully handle routing errors

mfenner:
## Purpose
The REST API should return a standard 404 Not Found message when an unknown route is called.

## Approach
Use a catch all route in the Router as the last step and raise `ActiveRecord::RecordNotFound`.


sarala:
@mfenner the solution is clear to me. Would this hide any serious errors? Why are we replacing ActionController::RoutingError with ActiveRecord::RecordNotFound? It feels like they are two different errors. May be I still don't understand the concepts very well.

PR #568, opened 2020-06-16
Set registration-agency

mfenner:
## Purpose
Set DOI registration agency for all DOIs using the lowercase id.

## Approach
To allow queries by DOI registration agency, we want to use a lowercase string, so `datacite` instead of `DataCite`, `crossref` instead of `Crossref` and `medra` instead of `mEDRA`. This pull request implements a rake task to update this information for all DOIs that need the update. It will also add `agency` information to those DOIs that don't have it.

## Learning
Using the `Doi.loop_through_dois` method greatly simplifies these kinds of metadata updates.

### DIFF
--- a/app/jobs/update_doi_job.rb
+++ b/app/jobs/update_doi_job.rb
@@ -1,0 +1,15 @@
+class UpdateDoiJob < ActiveJob::Base
+  queue_as :lupo_background
+
+  def perform(doi_id, options={})
+    doi = Doi.where(doi: doi_id).first
+
+    if doi.blank?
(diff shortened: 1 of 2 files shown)


mfenner:
@kjgarza the DOI update triggers this code that has a `before_save` callback (doi model, line 1907):
```
  def update_agency
    if agency.blank? || agency.casecmp?("datacite") 
      self.agency = "datacite"
    elsif agency.casecmp? ("crossref")
      self.agency = "crossref"
    end
  end
```

PR #575, opened 2020-06-29
Normalize Doi field language

mfenner:
## Purpose
Field is supposed to contain language in iso639-1 format, but frequently contains iso639-2 or English language version of the language. Normalization will help with faceting. We have 11 million DataCite DOIs with a language property, so more than 50%.

## Approach
Use iso639 ruby gem. Validation before saving language field. Will be used in GraphQL API and DataCite Commons.


mfenner:
Tests work, but needs to be resynced with master branch.

Issue #498, opened 2020-05-05
Identifiers returns null when data is available in alternateIdentifiers


richardhallett:
The "identifiers" attribute is serialised with "NULL" data regardless of what exists in the data for "alternate_identifiers", this also seems to be happening with the data in general with new DOI's.

Reproduce: Create a DOI using "alternateIdentifiers" to send the data.

```
{
  "data": {
    "attributes": {
      "doi": "10.70048/05052001",
      "alternateIdentifiers": [
        {
          "identifier": "10.70048/rh-0w-pi-58-b6-2",
          "identifierType": "DOI"
        }
      ]
    },
    "type": "dois"
  }
}
```

Look at response, notice NULLS.

Also another example where the "alternateIdentifiers" isn't NULL but "identifiers" is

Example DOI: https://api.test.datacite.org/dois/10.70048/rh-0w-pi-58-b6-2

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_2npud)

PR #583, opened 2020-07-16
Treat identifiers and alternate_identifiers as aliases. #498

mfenner:
## Purpose
Properly implement `identifiers` and `alternateIdentifiers` attributes.

## Approach
* `identifiers` should be a writable attribute in the REST API. This means the doi of the resource should not be included.
* this makes `identifiers` and `alternateIdentfiiers` aliases that should work the same way. This is implemented in the controller, before parameters are validated.
* we need to remove the dois from the `identifiers` attribute. This is done in the serializer for now, but there is also a rake task to learn up the data stored in the database. This needs to be documented and communicated.

## Learning
Alias attributes turned out to be a very painful activity, and there are several ways how this can be implemented. In the end the controller before checking strong parameters was the best place.


spec/requests/dois_spec.rb
```diff
@@ -3296,45 +3354,47 @@
...
-        {
-          "data" => {
-            "type" => "dois",
-            "attributes" => {
-              "geoLocations" => geo_locations
-            }
-          }
-        }
-      end
+    # TODO needs updating
```
richardhallett, line 3357: Is this broken due to the identifiers change? I can't quite see why this test would fail.

mfenner: I don't see a connection to changes in this pull request, that is why I commented it out. Need to look at it independently.

PR #584, opened 2020-07-16
Only trigger import message on changes in specific fields

mfenner:
## Purpose
Updates of DOI metadata should only trigger sending a message to `levriero` if the change is relevant for Event Data.

## Approach
* use `saved_change_to_attribute?("name")` method
* watch changes in `related_identifiers`, `creators`, `funding_references`

- [ ] this needs to be tested in AWS infrastructure including SQS


mfenner:
Expected functionality confirmed in our stage environment, as this can only be tested with AWS and not locally.

PR #587, opened 2020-07-20
Run loop_through_dois in background

mfenner:
## Purpose
Most rake tasks using the `loop_through_dois` method run for a long time, and we have to keep the server connection open. Sending the messages to a background queue earlier addresses this problem.

## Approach
The implementation is basically following the original proposal by @kjgarza.


mfenner:
The cursor generated by the options is broken (should be an array of timestamp and doi), and I removed it.

mfenner:
Thanks Kristian. Will look for test coverage of repository transfer going forward.

PR #588, opened 2020-07-22
Implement STI (single-table inheritance) for DOIs

mfenner:
## Purpose
Separate out non-DataCite DOIs in the model, database Elasticsearch index and APIs.

## Approach
Use Rails STI (single-table inheritance) to have multiple models and one database table. The models are Doi, DataciteDoi, and OtherDoi, and most of the functionality is in the Doi model. Elasticsearch doesn't fully work with STI, so we have to handle the Elasticsearch mappings, we do this by storing the mappings in Elasticsearch.

- [ ] More fine-tuning of using new Elasticsearch setup in tests.

## Learning
STI seems to be a good approach to the problem. Learned about Elasticsearch mappings. For changes to database tables we wait for Rails 6.1 with better support for multiple databases.


kjgarza:
I thought the objective of the work is to separate the data (i.e., Datacite, Other agencies). This change, it looks like, separates only the code using the STI pattern. Could we clarify that?

mfenner:
@kjgarza As I said, this pull request separates the model and Elasticsearch indexes. The database tables are not yet physically separated, and we use STI for logical separation. Changes in the database layout are much easier with the upcoming Rails 6.1, and will be added later, also we see the biggest impact in separating Elasticsearch indexes.

db/schema.rb
```diff
@@ -36,17 +36,12 @@
   create_table "allocator", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
     t.string "system_email", null: false
     t.datetime "created"
-    t.integer "doi_quota_allowed", null: false
-    t.integer "doi_quota_used", null: false
-    t.binary "is_active", limit: 1
+    t.binary "is_active", limit: 1, default: "b'1'"
```
kjgarza, line 39: I am trying to test this locally but it doesn't work. This error is both locally and in travis

> ActiveRecord::StatementInvalid: Mysql2::Error: Invalid default value for 'is_active':

mfenner: This is an issue unrelated to STI and I didn't notice when running locally. Will clean this up.

kjgarza:
@mfenner  There most be something else missing in the config, because also travis is passing even when Rspec is failing 

``` Finished in 23 minutes 0 seconds (files took 7.23 seconds to load)
1495 examples, 205 failures
Failed examples:
```

richardhallett:
While I think STI is an okay solution, and this implementation will probably be fine, I do worry it is complicating matters a little with the fact there are now multiple types only distinguished in ruby.
While it's not explicitly same situation this gitlab dont recommend using STI you can read more on why here: https://docs.gitlab.com/ee/development/single_table_inheritance.html 
The main concern would be if it diverges within the table and we have two types being represented on one table with different fields.

I guess it depends on how different the two types of DOIs we have, I know we convert into datacite metadata so it should be same.

I do wonder given we're having to do explicit mapping for creating the indexes if we couldnt just leave it as it was and determine the type by the "type".

Another option would be I guess to do more restructuring of the database, but as we saw this will need careful planning.  I'd like potentially something where we had the metadata seperated from the DOI, I think this would be easier to work with especially if we had a metadata table that was standardised across all doi types.

These are just some general thoughts, I'd need to spend a bit more time to give more concrete ideas.

In summary, the architecture I think maybe we could work differently, but as for this PR and its implementation I think it looks okay.

mfenner:
@richardhallett thanks for the helpful feedback. Some people don't like STI (including the maintainers of the elasticsearch-ruby gem) and it can turn very painful if the tables diverge too much. But we need (almost) exactly the same table. Only difference is currently that we are not using client_id/datacentre for Crossref DOIs, but I think that is acceptable.

I also agree that more work on restructuring the database would be helpful at some point, but I don't see any urgent need to do this for this particular work. 

We could do the separation also differently, but I hope that this is one acceptable implementation.

One of the things I want to do going forward is use the different DOI models to better implement the different functionalities of DataCite vs. other DOIs, e.g. whether registration in the handle system is possible.

mfenner:
I have merged the pull request after a lot of additional work on testing, e.g. not setting up alias indexes in testing. Significant further work will be needed before this can be deployed to production.

PR #595, opened 2020-08-03
Add migration for single-table-inheritance

mfenner:
## Purpose
Add migration needed by single-table-inheritance (STI) for the DOI model.

## Approach
This pull request separates the database migration from the application code changes, with the migration running first. This allows zero-downtime changes. The migration uses the `departure` gem, using `pt-online-schema-change` from the Percona Toolkit under the hood. This replaces the `lhm` gem that provides similar functionality, but is no longer actively maintained (last release in 2015). To avoid ssh connection errors we use `tmux`.


db/schema.rb
```diff
@@ -247,33 +247,6 @@
     t.index ["uuid"], name: "index_events_on_uuid", unique: true, length: 36
   end
 
-  create_table "lhmn_events", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin", force: :cascade do |t|
```
kjgarza, line 250: I am not sure I don't understand why is this tabled being removed. Could you elaborate?

db/migrate/20200313163242_rename_prefix_tables.rb
```diff
@@ -1,24 +1,22 @@
...
+    remove_column :prefix, :version, :integer
+    remove_column :prefix, :updated, :datetime
+    rename_table :prefix, :prefixes
 
-      rename_column :allocator_prefixes, :allocator, :provider_id
-      rename_column :allocator_prefixes, :prefixes, :prefix_id
-      add_column :allocator_prefixes, :uid, :string, null: false
-      rename_table :allocator_prefixes, :provider_prefixes
-      add_index :provider_prefixes, [:uid], length: 128
+    rename_column :allocator_prefixes, :allocator, :provider_id
```
kjgarza, line 9: When running the migration this fails:

```
ActiveRecord::StatementInvalid: Departure::Error: No foreign keys reference `datacite`.`allocator_prefixes`; ignoring --alter-foreign-keys-method.
The new table `datacite`.`_allocator_prefixes_new` does not have a PRIMARY KEY or a unique index which is required for the DELETE trigger.
: ALTER TABLE `allocator_prefixes` DROP PRIMARY KEY

```

I think its because there is no table `allocator_prefixes ` but I am not sure, what do you think?

mfenner: The `allocator_prefixes` table was renamed to `provider_prefixes` a few months ago.

db/schema.rb
mfenner: This is a temporary table generated by the `lhm` gem.

mfenner:
@kjgarza there is a single migration in this pull request, and that is exactly the same migration that was in the bigger pull request on implementing STI, i.e. the last migration in the master branch.

Issue #610, opened 2020-08-25
Create DMP type in GraphqL API


kjgarza:
To add a resource and queries for DMPs

Custom Aggregation for DMPs
> Query would be types.resourceTypeGeneral:Text AND types.resourceType:"Data Management Plan"

https://doi.datacite.org/dois?page=NaN&query=types.resourceTypeGeneral%3AText%20AND%20types.resourceType%3A%22Data%20Management%20Plan%22&resource-type-id=&sort=relevance

Respond to the query:

![carbon (24)](https://user-images.githubusercontent.com/1092861/91168564-21a3e380-e6d6-11ea-9dfe-76b924a5ad2c.png)

Aggregations:

- Datasets. Includes all citations and references of type dataset
- Funders. alias for fundingReferences
- organisations. Includes ror ids in `funderIndetifiers` and in creators.affiliation and contributor.affiliation if (contributor role is HostingInstitution)
- People. Includes all creators and contributors

mfenner:
@kjgarza I would recommend to not change the built-in types used elsewhere, e.g. what you propose with `funders` or `people`.

PR #614, opened 2020-08-28
Initial support for GraphQL data management plan type.

mfenner:
## Purpose
Add a data management plan type to the GraphQL API to specifically look for this information,

closes:  https://github.com/datacite/lupo/issues/610

## Approach
Initial implementation has the same functionality as other work types, e.g. `dissertation`.

## Types of changes

### DIFF
--- a/app/graphql/types/data_management_plan_type.rb
+++ b/app/graphql/types/data_management_plan_type.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class DataManagementPlanType < BaseObject
+  implements DoiItem
+
+  def type
+    "DataManagementPlan"
+  end
+end
--- a/app/graphql/schema.graphql
+++ b/app/graphql/schema.graphql
@@ -1778,6 +1778,292 @@ type DataCatalogEdge {
   node: DataCatalog
 }
 
+type DataManagementPlan implements DoiItem {
+  """
+  Metadata in bibtex format
+  """
+  bibtex: String!
+
+  """
+  Total number of citations
+  """
+  citationCount: Int
+
+  """
+  Citations for this DOI.
+  """
+  citations(affiliationId: String, after: String, fieldOfScience: String, first: Int = 25, funderId: String, hasCitations: Int, hasDownloads: Int, hasFunder: Boolean, hasOrganization: Boolean, hasParts: Int, hasPerson: Boolean, hasVersions: Int, hasViews: Int, ids: [String!], language: String, license: String, memberId: String, published: String, query: String, registrationAgency: String, repositoryId: String, resourceTypeId: String, userId: String): WorkConnectionWithTotal
+
+  """
+  Citations by year
+  """
+  citationsOverTime: [YearTotal!]
+
+  """
+  The container (e.g. journal or repository) hosting the resource.
+  """
+  container: Container
+
+  """
+  The main researchers involved in producing the data, or the authors of the publication, in priority order
+  """
+  creators(first: Int = 20): [Creator!]
+
+  """
+  Different dates relevant to the work
+  """
+  dates: [Date!]
+
+  """
+  All additional information that does not fit in any of the other categories
+  """
+  descriptions(first: Int = 5): [Description!]
+
+  """
+  The DOI for the resource.
+  """
+  doi: String!
+
+  """
+  Total number of downloads
+  """
+  downloadCount: Int
+
+  """
+  Downloads by month
+  """
+  downloadsOverTime: [YearMonthTotal!]
+
+  """
+  OECD Fields of Science of the resource
+  """
+  fieldsOfScience: [FieldOfScience!]
+
+  """
+  Technical format of the resource
+  """
+  formats: [String!]
+
+  """
+  Metadata as formatted citation
+  """
+  formattedCitation(locale: String = "en-US", style: String = "apa"): String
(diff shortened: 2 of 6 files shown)


app/graphql/types/data_management_plan_type.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class DataManagementPlanType < BaseObject
+  implements DoiItem
+
```
kjgarza, line 5: What about adding a  `datasets` field that aggregates DoiItem **citations** and **references** together but filters them by type "**Dataset**". This would remove significant processing for the user of the DMP which would be looking for datasets related to the DMP and not DOIs related. That aligns, IMO, closer to the vision of the DMP.

app/graphql/types/data_management_plan_type.rb
```diff
@@ -0,0 +1,9 @@
...
+  def type
```
kjgarza, line 6: What about adding a `organisations` field, that aggregates all the affiliations found in both creators and contributors and those affiliations from contributors HostingInstitution. In the vision of the DMP mentions that we want the organisations, with a field as such users of the API will have it right away.

kjgarza (commented):
I think we can add some extra field in the DMP type to make this more covenant for users. See comments in the review.

app/graphql/types/data_management_plan_type.rb
mfenner: I would not make these kinds of changes only for DMPs. I want the GraphQL API to be as generic as possible.

app/graphql/types/data_management_plan_type.rb
mfenner: If possible I would avoid any functionality that is specific to DMPs. It makes it more complicated for users to use the API, and I assume we need to refactor when we release the next schema with support for ResourceTypeGeneral `dataManagementPlan`.

What one can do instead to customize the query is using an [inline fragment](https://graphql.org/learn/schema/):

```
query HeroForEpisode($ep: Episode!) {
  hero(episode: $ep) {
    name
    ... on Droid {
      primaryFunction
    }
  }
}
```

kjgarza: > It makes it more complicated for users to use the API,

Can you elaborate on this? I do not think I understand why adding functionality to a more specific type would make this more complicated for users. Particularly when they will have to scan the response to find what they actually need. The use case of the DMPs talks about, exclusively,  `datasets`. It feels to me that without adding this functionality in the type, we are making the user to work more and to have knowledge about what constitute a `datasets` from all the metadata response. 

I would like to understand why is this more complicated.

app/graphql/types/data_management_plan_type.rb
kjgarza: Would this mean that we would let the user to figure out what constitute an `organisation` and/or `dataset` out of all the metadata they get in query for DMP (basically `OverFetching`)?

app/graphql/types/data_management_plan_type.rb
mfenner: The design of the DataCite GraphQL API tries to be as generic as possible, e.g. in terms of supported fields. Assuming that users will not only query for DMPs, or at least they may reuse pre-existing notebooks and tweak them. Please look into the suggestions I made regarding the query on the client side.

app/graphql/types/data_management_plan_type.rb
mfenner: Of course not. Organizations and datasets are different fields, so they can decide what they want to fetch.

kjgarza:
I would have expected users to make a query like this to get all the datasets of a DMP:

```
{
  datamanagementplan(id: "10.25495/xkya-0g76") {
      id
      datasets {
        totalCount
        nodes {
          id
          titles {
            title
          }
        }
      }
  }
}

```



instead, they will have to do something like 



```
{
  datamanagementplan(id: "10.25495/xkya-0g76") {
      id
      references(resource_type_id:"dataset") {
          totalCount
          nodes {
            doi
            titles {
              title
            }
          }
      }
      citations(resource_type_id:"dataset") {
          totalCount
          nodes {
            doi
            titles {
              title
            }
          }
      }
  }
}

```

and then merge the response. is that the case?

mfenner:
Again, please read my comment regarding aliases and inline fragments.

kjgarza:
@mfenner  alright, this is what I refer about the letting the user figure it out: that **references filtered by dataset type**  = **datasets**.  

Ok, so the query for datasets is not as complicated as I thought (still not hat convenient). 





What about **organisations** then? should it be something like this

```
{
  dataManagementPlan(id: "10.25495/xkya-0g76") {
      id
    fundingReferences{
      funderIdentifier
      funderIdentifierType
    }
    contributors{
      id
      contributorType
    }
   }
}

```

and then the **user** will need to (on the client side): 

1. filter the **contributor by Type**.  Because **contributors** has no filters to do `contributors(contributorType:"HostingInstituion"){`
2. and then filter also the `fundingReferences` by those ones with ror_ids 
3. and then merge both of them

is this the intention?

mfenner:
There is more than one way to do this. For the time being funding information will almost always use Crossref Funder IDs. For filtering by ContributorType, I suggest adding a filter. #615 is not merged yet.

PR #627, opened 2020-09-15
Add rake task to trigger updatedois based on query

richardhallett:
## Purpose
Generic update dois rake task based on supplied query.

After looking at how the existing rake tasks work and what we need to do to fix missing data, triggering an update based on a query would be the easiest.
This generic allows us to 
a) do specific DOIs if need be
b) use things like -agency:* - To search for empty agency

## Approach
New take task based on existing

N/A
## Learning
Applied locally and checked behaviour


kjgarza (approved):
LGTM

PR #631, opened 2020-09-16
Use cursor when supplied via options

richardhallett:
Without this the cursor always starts off at []

## Purpose
Fix cursor nav in update dois task


richardhallett:
Not needed. base64 encode without the list part []

PR #633, opened 2020-09-16
Add ES request_timeout env var

richardhallett:
This is to set a slightly higher default and allow override on ENV.

See https://www.rubydoc.info/gems/elasticsearch-transport#Setting_Timeouts for more info.


mfenner (approved):
LGTM

PR #645, opened 2020-09-28
Add rake task for fixing index type

richardhallett:
Add rake task for fixing the type column on datasets

## Purpose
This rake task adds the type information either DataCite or OtherType based on agency.

## Approach
Rake task

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -868,7 +868,7 @@ class Doi < ActiveRecord::Base
     filter << { terms: { "client.certificate" => options[:certificate].split(",") }} if options[:certificate].present?
     filter << { term: { "creators.nameIdentifiers.nameIdentifier" => "https://orcid.org/#{orcid_from_url(options[:user_id])}" }} if options[:user_id].present?
     filter << { term: { "creators.nameIdentifiers.nameIdentifierScheme" => "ORCID" }} if options[:has_person].present?
-    
+
     # match either one of has_affiliation, has_organization, has_funder or has_member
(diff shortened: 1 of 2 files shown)


app/models/doi.rb
```diff
@@ -2310,4 +2310,39 @@ def self.migrate_landing_page(options={})
...
+
+    Doi.where(id: from_id..until_id).where('type' => nil).find_each(batch_size: 500) do |doi|
+      begin
+        if doi.agency.casecmp?("datacite")
+          type = "DataciteDoi"
+        else
+          type = "OtherDoi"
+        end
+
+        doi.update_columns("type": type)
```
mfenner, line 2334: This line should either be `(type: type)` or `("type" => type)`.

app/models/doi.rb
```diff
@@ -2310,4 +2310,39 @@ def self.migrate_landing_page(options={})
...
+    until_id = (options[:until_id] || (from_id + 499)).to_i
+
+    # get every id between from_id and end_id
+    count = 0
+
+    Rails.logger.info "[migration_index_types] adding type information for DOIs with IDs #{from_id} - #{until_id}."
+
+    Doi.where(id: from_id..until_id).where('type' => nil).find_each(batch_size: 500) do |doi|
+      begin
+        if doi.agency.casecmp?("datacite")
```
mfenner, line 2328: You want to handle the case of agency.blank?

mfenner (approved):
Two minor comments.

PR #646, opened 2020-09-28
better error handling

kjgarza:
## Purpose
Better handling of array validation
closes: https://github.com/datacite/lupo/issues/643

## Types of changes

### DIFF
--- a/app/controllers/datacite_dois_controller.rb
+++ b/app/controllers/datacite_dois_controller.rb
@@ -666,10 +666,12 @@ class DataciteDoisController < ApplicationController
       xml = xml.strip
     end
 
+    fail(ActionController::UnpermittedParameters, ["creators must be an Array"]) if p[:creators]&.respond_to?(:keys)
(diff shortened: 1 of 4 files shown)


richardhallett (approved):

mfenner:
@kjgarza I would have like to give feedback on this one. Why do you raise an error if creators are not an array? You could instead made sure that the input is transformed into an array as needed, as we do in a number of places.

kjgarza:
Basically so users send the payload always in the correct schema format [(:array)](https://github.com/datacite/schema/blob/6ad7d7005d03b35e136fc4d8cd77a8e4f1045e72/source/json/kernel-4.3/datacite_4.3_schema.json#L301). In terms of the approach, we have follow this same error raising before in https://github.com/datacite/lupo/pull/591. One could say neither of us was aware that the API should be sanitising users' payloads in the way you are suggesting.

mfenner:
Please change this code to transform to array rather than raise an error.

Issue #641, opened 2020-09-23
Metadata transformation problem for alternateIdentifiers


kjgarza:
this could be  an API problem, like the API is not converting alternativeIDentifiers (XML) to identifiers (JSON) the metadata in json doesn't show them either https://api.datacite.org/dois/10.24397/pangloss-0000008

<img src="https://frontapp.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_2o3px)

kjgarza:
For context, Bolognese Transformation is working correctly with this metadata file

kjgarza:
The problem is related to way we replace the `alternatedIdentifiers` with `identifiers`.  

When sending XML the identifier field is `blank` but is not `nil` therefore the API takes it (see [lupo:L147](https://github.com/datacite/lupo/blob/a85b412/app/controllers/datacite_dois_controller.rb#L714) ) as if the request included and empty value.  This condition is not necessarily wrong as API might want to replace a value with `blank`.

richardhallett:
I can reproduce this.

1. Create a XML document with alternateIdentifiers listed
2. Do a GET from https://api.stage.datacite.org/dois/thedoi
3. Note the response does not contain the alternateIdentifiers in either identifiers or alternateIdentifiers field.
4. Look at XML note it does contain the alternateIdentifiers

Example XML
```
<?xml version="1.0" encoding="UTF-8"?>
<resource xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://datacite.org/schema/kernel-4" xsi:schemaLocation="http://datacite.org/schema/kernel-4 http://schema.datacite.org/meta/kernel-4/metadata.xsd">
  <identifier identifierType="DOI">10.24413/DMVJ-JS47</identifier>
  <creators>
    <creator>
      <creatorName nameType="Personal">Hallett, Richard</creatorName>
      <givenName>Richard</givenName>
      <familyName>Hallett</familyName>
    </creator>
  </creators>
  <titles>
    <title>Title</title>
  </titles>
  <publisher>Acme Inc</publisher>
  <publicationYear>2020</publicationYear>
  <resourceType resourceTypeGeneral="Audiovisual">MediaObject</resourceType>
  <alternateIdentifiers>
[...]

PR #650, opened 2020-10-07
Refactor strong params to deal with replacement of values

kjgarza:
## Purpose

There is a problem in that replace the `alternatedIdentifiers` with `identifiers`.  

When sending XML the identifier field is `blank` but is not `nil` therefore the API takes it (see [lupo:L147](https://github.com/datacite/lupo/blob/a85b412/app/controllers/datacite_dois_controller.rb#L714) ) as if the request included an empty value.  This condition is not necessarily wrong as API might want to replace a value with `blank`.

closes: https://github.com/datacite/lupo/issues/641

## Approach

conditioning of which parameters (json params or xml params) are accepted from the API is complex and by the look at the problem of  `alternatedIdentifiers`  inconsistent.

Here I opted for refactoring `strong_params` to accept parameters in separated methods.

## Learning

We accept data even when is contradictory:

```
      {
          "data" => {
            "type" => "dois",
            "attributes" => {
              "url" => "http://www.bl.uk/pdf/pat.pdf",
              "xml" => xml,
              "titles" => "Submitted chemical data for InChIKey=YAPQBXQYLJRXSA-UHFFFAOYSA-N",
              "event" => "publish"
            }
          }
        }

```

this payload can include different values for the same parameters.

## Types of changes

### DIFF
--- a/app/controllers/datacite_dois_controller.rb
+++ b/app/controllers/datacite_dois_controller.rb
@@ -666,12 +666,28 @@ class DataciteDoisController < ApplicationController
     #   xml = xml.strip
     # end
 
-    p = case p[:source]
-        when "fabrica" || "fabricaForm" || "mds" || "ezid" || "legacy"
-          xml_params(p)
-        else
-          json_params(p)
-        end
+    read_attrs_keys = [:url, :creators, :contributors, :titles, :publisher,
+      :publicationYear, :types, :descriptions, :container, :sizes,
+      :formats, :language, :dates, :relatedIdentifiers,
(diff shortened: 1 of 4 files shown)


kjgarza:
- Do you think this is an approach we should be pursuing ?
- Do you see/know any edge cases for metadata submission which would be problematic if we continue with this approach?



@mfenner @richardhallett

richardhallett:
This was solved with a bug fix, see commit 
https://github.com/datacite/lupo/commit/b715a914f8f373abb1f99747644c25b991618ff6

I think given the age of this PR it's best to close for now and we revisit if we need to. 
Thanks.

PR #656, opened 2020-10-15
sanitize wrong metadata

kjgarza:
addresses: https://github.com/datacite/lupo/issues/655

## Purpose

## Types of changes


app/controllers/datacite_dois_controller.rb
```diff
@@ -666,13 +666,11 @@ def safe_params
       xml = xml.strip
     end
 
-    fail(ActionController::UnpermittedParameters, ["creators must be an Array"]) if p[:creators]&.respond_to?(:keys)
-    p[:creators]&.each do |c|
+    Array.wrap(params[:creators])&.each do |c|
```
richardhallett, line 669: So we're going to allow creators to be sent as not an array? isn't that bad, shouldnt we require them to send data correctly rather than try and "fix" it?

mfenner: I am all for avoiding raising errors, in particular if the fix is as simple as this one. At some point we need to refactor our code regarding "forcing into an array", as we do this in too many different places (e.g. also model and serializer).

richardhallett: But these are user errors and therefore shouldnt they be raised because it should be to encourage proper usage of the API?
What is perhaps a related problem is the fact these went to sentry at all, these are not application errors as nothing is wrong here really, it was just bad data, so there is nothing to fix, or at least that's how I see it.
But maybe that's too big a fix for this particular issue.

kjgarza: I do agree with Richard, I think that telling users that their data is formatted incorrectly (wrong types) is a better approach. It keeps their implementation safe and also it helps technical support because one doesn't have to be guess whether some data submission was sanitised or not.

mfenner: As a general principle I like to be as flexible as possible for input and as stringent as possible for output. In this particular case there are three specific issues:

* this relates to how we have implemented our metadata in XML and JSON, and we are not always consistent with regards to arrays.
* schema.org expects arrays with one element to be an object instead
* we are forcing objects to be an array in many places in our codebase

Also, this implementation is simple and safe.

richardhallett (approved):
Approving basis of conversation that this is an acceptable solution.
I personally feel that we should be handling these things different, the API should be stricter, but that's a bigger conversation than this specific PR.

mfenner (approved):
LGTM

PR #665, opened 2020-10-22
Add success check after DB update for ES indexing

richardhallett:
Only do an ES index if we successfully saved to the DB with correct information.

## Purpose
To fix accidental double indexing when record is in bad state:
https://github.com/datacite/datacite/issues/1134

## Approach
Just check to see if the save passed or not.

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/app/models/concerns/helpable.rb
+++ b/app/models/concerns/helpable.rb
@@ -21,10 +21,6 @@ module Helpable
         raise ActionController::BadRequest.new(), "[Handle] Error updating DOI " + doi + ": client ID missing."
       end
 
-      unless match_url_with_domains(url: url, domains: client.domains)
-        raise ActionController::BadRequest.new(), "[Handle] Error updating DOI " + doi + ": URL not allowed by repository domains settings."
-      end
-
       unless is_registered_or_findable?
         raise ActionController::BadRequest.new(), "DOI is not registered or findable."
       end
--- a/app/models/concerns/helpable.rb
+++ b/app/models/concerns/helpable.rb
@@ -57,6 +57,7 @@ module Helpable
 
       if [200, 201].include?(response.status)
         # update minted column after first successful registration in handle system
+        success = true
         success = self.update_attributes(minted: Time.zone.now, updated: Time.zone.now) if minted.blank?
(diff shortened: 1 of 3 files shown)


mfenner (commented):
I think I would slightly rewrite as success is not always defined. 

But something is not right in the flow. I was struggling with this when I initially wrote this, maybe it makes sense to look at the Rails callbacks again: https://guides.rubyonrails.org/active_record_callbacks.html

richardhallett:
Ah I assumed that it always returned a boolean value.
Okay added default check.

As for the callbacks, this is what I was saying in the issue, it was changed to before_save, when it used to be after commit? I don't quite understand your thinking here, can we not just take it back to after commit? You can still presumably leave it doing the register_url directly rather than a background job

mfenner:
@richardhallett what I am not 100% sure is whether an after_save or after_commit happens before returning something to the controller. And not asynchronously.

richardhallett:
I've updated to make it return 422 for the domain specific problem.
Other URL validation already occurs, so now the handle registration will just log again specific handle problems.

app/models/concerns/helpable.rb
```diff
@@ -57,10 +53,13 @@ def register_url
 
       if [200, 201].include?(response.status)
         # update minted column after first successful registration in handle system
-        self.update_attributes(minted: Time.zone.now, updated: Time.zone.now) if minted.blank?
+        success = true
```
mfenner, line 56: shouldn't this be `success = false`?

app/models/doi.rb
```diff
@@ -1836,6 +1837,12 @@ def event=(value)
     self.send(value) if %w(register publish hide show).include?(value)
   end
 
+  def check_url
```
mfenner, line 1840: I like this approach, makes it much easier. You should check whether the the DOI is in draft state, when any DOI should be allowed.

mfenner (approved):
Looks good, only two minor comments.

app/models/concerns/helpable.rb
richardhallett: No because as far as I can tell the update_attribute will only return false if there is a problem therfore the default is passing unless it tells you otherwise.

PR #680, opened 2020-11-19
Count dois not indexed by repository

mfenner:
## Purpose
Generate a listing of DOIs not indexed by repositories.

## Approach
Loop through all repositories and count number of DOIs in db and Elasticsearch. List repository if there is a difference with total number of dois and number missing.

## Types of changes

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -448,6 +448,73 @@ class Client < ActiveRecord::Base
     { "id" => symbol.downcase, "type" => "clients", "attributes" => attributes }
   end
 
+  def self.export_doi_counts(query: nil)
+    begin
+      # Loop through all clients
+      page = { size: 1000, number: 1 }
+      response = self.query(query, page: page)
+      clients = response.results.to_a
+
+      total = response.results.total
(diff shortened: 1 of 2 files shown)


mfenner:
The commit https://github.com/datacite/lupo/commit/b16a04577b0313c9ed2b38995d5e872cbc67a06c didn't make it into the pull request, but includes the rake task that triggers the report generation.

richardhallett (approved):
Looks fine, will be quite slow doing a individual query for each client count, but without a different index on the DB this probably is the easiest route.

mfenner:
@richardhallett I briefly looked into MySQL `group by`. This can dramatically cut down on the number of queries, but be a very big query.

richardhallett:
> @richardhallett I briefly looked into MySQL `group by`. This can dramatically cut down on the number of queries, but be a very big query.

A group by would be the canonical and probably preferred way to do this, saves the IO round trip, and SQL is better at grouping, however without an index on the datacentre its probably not going to be that efficient given size of our DB:)

PR #684, opened 2020-12-03
Prefix transfer fix

richardhallett:
This is effectively a revert of PR #642
the reasoning is this caused subtle bugs with
model changes that did not go via a controller.

The problem this caused was that the database would be out of sync
with the index.

This was most notable in repository transfers where the DB
would show correct transfers but index remained unchanged.

## Purpose
To fix transfers so prefixes are correctly indexed.

## Approach
Core is that we only want to change the ES index when model changes happen, not only on controllers.

However to accommodate the original idea of the previous commits, this will now immediately do a index job on the model changes rather than wait on a queue.

## Post merge
Test on staging to ensure transfers behave correctly.

## Types of changes

### DIFF
--- a/app/controllers/client_prefixes_controller.rb
+++ b/app/controllers/client_prefixes_controller.rb
@@ -128,27 +128,6 @@ class ClientPrefixesController < ApplicationController
     authorize! :create, @client_prefix
 
     if @client_prefix.save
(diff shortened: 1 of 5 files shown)


kjgarza (approved):
utACK

AFAIK this should probably fix it. I cannot tested locally as the setup rake tasks are broken.

PR #704, opened 2021-01-12
Fix provider prefix

richardhallett:
## Purpose
Fixes the 422 error thrown on assigning a prefix

## Approach
Correct check in controller

## Types of changes

### DIFF
--- a/app/controllers/provider_prefixes_controller.rb
+++ b/app/controllers/provider_prefixes_controller.rb
@@ -125,9 +125,7 @@ class ProviderPrefixesController < ApplicationController
     @provider_prefix = ProviderPrefix.new(safe_params)
     authorize! :create, @provider_prefix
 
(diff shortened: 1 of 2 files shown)


mfenner (commented):
LGTM. Where do you trigger the indexing for the prefix?

richardhallett:
> LGTM. Where do you trigger the indexing for the prefix?

In the model, specifically in indexable concern. This was handled in the PR before Christmas see https://github.com/datacite/lupo/pull/684 for more details.
This was approved by Kristian due to your absence and it related to the previous transfer work he'd done.

So this PR is basically an addition to that one, it matches the way repository prefix assignment worked.

I actually don't think this is the root cause of the prefixes index not being updated however, but it will at least fix the error thrown when assigning.

mfenner (approved):
LGTM. Thanks for the clarification.

Issue #709, opened 2021-01-27
Optionally disable facets in REST API


mfenner:
By adding the optional parameter `facet-count` we can both disable facets in the DOI REST API to increase performance, or change the number of facets from the default `10`. We did the same change in the GraphQL API a few weeks ago (#701).

Issue #701, opened 2021-01-05
Option for GraphQL API to return all subject areas


antonninkov:
We would like to have the option of using the GraphQL API to return all the subject areas in the database. Currently, only 10 are returned. For example, in the following GraphQL call, we would like to see any subject areas/fieldsOfScience/disciplines that have datasets published. Right now, only the 10 subject areas with the most datasets are returned. Thanks!

`
{
  datasets {
    totalCount
    fieldsOfScience{
      title
      count
    }
}
}`

mfenner:
@antonninkov would something like this work?

```
{ datasets { 
    totalCount 
    fieldsOfScience(facetCount: 50) { 
      title 
      count
    }
  }
}
```

Where `facetCount` would have a default value of `10`, not breaking the current implementation.

Facets (Elasticsearch aggregations) are a major performance bottleneck, so this expanded result set will take some time. Probably not a problem in a Jupyter notebook, but this change would be noticeable in a web interface with live data.

antonninkov:
Yes, a solution like this would be great. It would also be great if this could similarly be applied to Licenses.

mfenner:
@antonninkov The following now works and should give you what you need:

```
{ datasets(facetCount: 50) {
    totalCount
    fieldsOfScience { 
      title 
      count
    }
  }
}
```

The facet count is provided at the top level, as that was the best way to implement this. This is an optional setting that defaults to 10, and also works for licenses or other facets, e.g. 

```
{ datasets(facetCount: 50) {
    totalCount
    licenses { 
      title 
      count
    }
  }
}
```

The performance is reasonable when using only a single facet like in the examples above. Let me know if that doesn't solve your issue.

antonninkov:
Thanks Martin. Would this all work for "published" variable?

mfenner:
No. That is currently hard-coded somewhere else, as faceting time-series works differently. Currently it is all years starting with 2010 and ending with the current year. I can include that, but maybe you want something else: start year and end year of the facets?

PR #710, opened 2021-01-27
Optionally change number of facets in DOI REST API

mfenner:
## Purpose
Change the number of facets in the DOI REST API from the default 10 to a different number or to 0. This is a global setting for all facets. Disabling facets will significantly improve performance.

closes: #709

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -585,78 +585,80 @@ class Doi < ApplicationRecord
     end
   end
 
-  def self.query_aggregations
-    {
-      resource_types: { terms: { field: "resource_type_id_and_name", size: 16, min_doc_count: 1 } },
-      states: { terms: { field: "aasm_state", size: 3, min_doc_count: 1 } },
-      published: {
-        date_histogram: {
-          field: "publication_year",
-          interval: "year",
-          format: "year",
-          order: {
-            _key: "desc",
+  def self.query_aggregations(facet_count: 10)
+    if facet_count.to_i.positive?
+      {
+        resource_types: { terms: { field: "resource_type_id_and_name", size: facet_count, min_doc_count: 1 } },
+        states: { terms: { field: "aasm_state", size: 3, min_doc_count: 1 } },
+        published: {
+          date_histogram: {
+            field: "publication_year",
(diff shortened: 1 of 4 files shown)


app/models/doi.rb
```diff
@@ -585,78 +585,80 @@ def self.gql_query_aggregations(facet_count: 10)
...
-        date_histogram: {
-          field: "publication_year",
-          interval: "year",
-          format: "year",
-          order: {
-            _key: "desc",
+  def self.query_aggregations(facet_count: 10)
+    if facet_count.to_i.positive?
+      {
+        resource_types: { terms: { field: "resource_type_id_and_name", size: facet_count, min_doc_count: 1 } },
```
richardhallett, line 591: With a global facet count, this will now return different number of resource types, previously 16. Do we have anything that depends on this?

richardhallett (commented):
As per the comment about facet_size being different for one.
I wonder if it's not better just to have a global disable flag rather than changing the facet count.
The size parameter on terms is something that can affect performance, the higher the required number of terms the more performance cost, so its potentially more dangerous, additionally it might be an area where we want to lower for specific aggregations.

app/models/doi.rb
mfenner: Good point. We could hardcode the actual number of possible facets for resource type if facet_count > 0.

mfenner: The reason I use an integer for the fact count instead of a boolean is that the initial issue for GraphQL (#701) had a different purpose, increasing the number of facets. In GraphQL you can show only the facet(s) you want giving you more flexibility.

Do you suggest that we don't allow that flexibility with the REST API, just having facets on or off?

richardhallett: Ah I see. Ye I think we should reduce the flexibility for the REST API, it makes things simpler.

mfenner: Ok, will do that.

mfenner:
@richardhallett I made the suggested change, now using the optional `disable-facets=true` query parameter.

richardhallett (approved):
lgtm

PR #711, opened 2021-02-01
Add delete by doi method

richardhallett:
## Purpose
Need a way to delete DOIS for work like: https://github.com/datacite/bracco/issues/502

## Approach
Use same delete job method as existing delete by prefixes

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -1855,6 +1855,13 @@ class Doi < ApplicationRecord
     errors.add(:language, "Language #{language} not found.") if entry.blank?
   end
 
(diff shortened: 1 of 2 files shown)


mfenner (commented):
How is this method triggered? I assume a rake task, and I would not enable this via API calls, even if only allowed for staff, as this should be a limited use admin task.

richardhallett:
Ah I accidentally missed the rake code commit. I will add but yes rake only to work like the delete_by_prefix

mfenner (approved):
LGTM

Issue #712, opened 2021-02-03
Increase request entity size


timrobertson100:
[GBIF](https://www.gbif.org) are receiving  "413 Request Entity Too Large" from DataCite when creating DOIs.

Our metadata contain relationships to other datasets and are often large as we can link to ~16,000 other datasets. We expect this to continue to grow.

Being able to correctly attribute and link the source datasets is critical to our operations.

Can you please increase this limit (suggest doubling it)?

mfenner:
@timrobertson100 we currently limit the request size to 10 MB via the nginx web server configuration. What is the file size of those DOIs with 16k related identifiers?

I assume you compress those requests with gzip before sending to us.

mike-podolskiy90:
@mfenner the size is about 15 MB

mfenner:
Do you expect this to be good enough for the next 2-3 years? Would increasing the file size limit 20 MB be sufficient? And could you share an example metadata record (ideally via URL) with that size?

timrobertson100:
@mike-podolskiy90 - can you please explore if GBIF are indeed compressing before sending to DataCite?

mike-podolskiy90:
We're not compressing files, so we gonna do that.
@mfenner thank you for the quick reply
Close this.

mfenner:
Excellent, compression should give you a lot of extra mileage for the 10 MB that are currently allowed. Ping us if that is not enough.

timrobertson100:
Will do. Thanks for pointing us in the right direction @mfenner

mfenner:
@timrobertson100 one more question: do you use the MDS API or the REST API? As the MDS API is, in turn, using the REST API, so two layers that each need to decompress and send forward.

timrobertson100:
REST API using our [Java Data Cite client](https://github.com/gbif/datacite-rest-client/blob/master/datacite-rest-client/src/main/java/org/gbif/datacite/rest/client/retrofit/DataCiteRetrofitClient.java)

This implements the parts GBIF need but may  be a good candidate for a more complete Java client. We'd be happy to move it to the DataCite repository if desirable, or see it forked.

mfenner:
Thanks Tim. We should have a discussion about how to make existing DataCite client applications more visible.

mike-podolskiy90:
@mfenner Sorry but I did not manage to perform a request with compressed data. And I also did not find a documentation how to do that. Could you provide me with a curl + payload example please?

mfenner:
[...]

PR #714, opened 2021-02-05
Issue #712 - Increase request entity size. (Allow for handling of com…

svogt0511:
## Purpose

closes: Issue #712.  Allows the rest api to accept compressed requests (gzipped body).  It seemed like that was not being handled currently.

## Approach

I added the same middleware that sashimi uses to do this.  Please check this over and let me know if:

- There was some other method I should have used.
- I need to make sure this is a general enough approach.

Definitely needs more testing before deployment.

## Types of changes

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -85,6 +85,9 @@ module Lupo
     config.paths.add Rails.root.join("app", "serializers", "v3").to_s,
     eager_load: true
 
+    # Allow middleware to be loaded. (compressed_requests)
+    config.autoload_paths += %W(#{config.root}/lib #{config.root}/lib/middleware)
+
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
--- a/config/initializers/middleware.rb
+++ b/config/initializers/middleware.rb
@@ -1,0 +1,5 @@
+require 'compressed_requests'
+
+Rails.application.configure do
+  config.middleware.insert_before Rack::Head, CompressedRequests
+end
--- a/config/initializers/middleware.rb
+++ b/config/initializers/middleware.rb
@@ -1,4 +1,6 @@
-require 'compressed_requests'
+# frozen_string_literal: true
+
+require "compressed_requests"
 
 Rails.application.configure do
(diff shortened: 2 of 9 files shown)


mfenner (commented):
I thought we handle the compression in nginx rather than Rails?

svogt0511:
@mfenner Good question.  I will check further to see what is happening. Currently the request body is coming through zipped.

svogt0511:
@mfenner, looking at the configuration (in my local dev) I see 'gzip on;', however, if I am not mistaken, the documentation says that applies to compressing responses, not requests.  We can talk about it on Monday.

kjgarza:
Sorry I was not following this conversation during the week. But yeah, this is correct. If the were to send gzip compress payload they would need to send a compressed string when using middleware.

mfenner:
Nginx supports decompression, see https://docs.nginx.com/nginx/admin-guide/web-server/compression/. Not sure whether using Rails middleware or Nginx for decompression is preferable. We have looked into nginx vs. application server in other cases, e.g. CORS.

svogt0511:
@mfenner I saw that - support of decompression using lua.  I was looking into adding the lua shared object module and integrating a script for that purpose.  We will see how that goes.

mfenner:
Keep in mind that we use a pre-built binary for nginx that includes passenger support. Adding additional nginx modules is extra work. But Openresty is a very interesting platform for other reasons, and also the Kong built on top of it.

mfenner (commented):
Please align with how we are including other middleware in config/application.rb.

Also, do we need a subfolder `middleware` in `lib`? It makes loading more complex.

svogt0511:
Working on writing some rspec tests for this.

svogt0511:
This seems to be working.  I have a test in postman.   I am working on that.  Also, I found that file submitted as a test - request-3.json.gz, was invalid, which is the reason for the error: "You need to provide a payload following the JSONAPI spec".  One small change and that worked.

svogt0511:
@mfenner - could you review and approve this when you get a chance?  It is finally ready to be pushed to production.  I added an rspec test which it passes.  I have also tested it with curl and that seems to work too.  Thanks!

mfenner (commented):
LGTM

mfenner (approved):
LGTM

Issue #716, opened 2021-02-10
Initial setup for version 3 REST API


mfenner:
A new REST API version is needed for the changes in schema 4.4, and highly desirable for the new contacts API. This issue described the effort for the initial setup, using a path-based versioning approach.

PR #719, opened 2021-03-09
Make facets for v3 events API opt-in

mfenner:
## Purpose
Elasticsearch aggregations cause a significant performance hit in API endpoints with many records, i.e. `/dois` and `/events`. In the upcoming v3 API we should theremore make them opt-in, rather than opt-out, as in the v2 API `/dois` endpoint. This is a breaking change.

closes: #716

## Approach
We are by default not showing any aggregations for `/events`. Each facet has to be specifically enabled with the `facet` query parameter, more than one can be separated by a comma, e.g. `&facet=sources,prefixes`. A similar approach could be taken for `/dois`.

The pull request also sets up initial request specs for the v3 API, copied from the existing request specs.

* do we want a query parameter that enables all aggregations at once, e.g. `&facet=all`?
* do we want to set the number of aggregations per facet via a query parameter? Crossref is doing `&facet=sources:20`.

## Types of changes


richardhallett:
I think in principle looks fine, we discussed seperatly concerns about code duplication but as a first draft of a new API the approach seems fine.
I think  your specific questions, I am inclined to say that we could allow those extra options via query string and yes then the performance is offloaded to the user, but is by default disabled.

So doing this for dois would also make sense.

richardhallett (approved):

PR #743, opened 2021-07-09
First upload of python scripts for posting data

kelvinlyy:
The scripts aim to automate XML files processing for REST API so that publishers can use DataCite's REST API for uploading metadata of publications in XML format


kjgarza:
I think this was a missunderstading @kelvinlyy . you need to create a totally new repository in https://github.com/kelvinlyy and save there only the files of your implementation

Issue #741, opened 2021-06-25
Improve GraphQL performance for Researcher Profile Pages (/orcid.org/{orcid}) 


kjgarza:
### Expected Behaviour

The researcher profile page response should keep the user's attention focused.

### Current Behaviour

The response time of the researcher profile pages is very slow and it takes a long time. This was identified as feedback from [AGU and PARSEC](https://datacite.productboard.com/roadmap/2537097-development-journey/notes/16578737). 

Current LCP (Largest Contentful Paint) for the /orcid.org/{orcid} varies significantly but the research profile page with related works has response times [above >10sec](https://docs.google.com/document/d/1mhOm-LassOlPGE2t3_TwjENFHI4CApZaqH5R7sYjDKs/edit#bookmark=id.dxgki8fbduox). 

![image](https://user-images.githubusercontent.com/1092861/123425946-08ecae80-d5c3-11eb-9b81-8de0eea52ef8.png)

### Possible Implementation

As a user with works, I expect the profile page (/orcid.org/{orcid}) response time to be under 10 seconds (LCP < 10s for desktop at P90)

PR #756, opened 2021-09-07
Cache graphQL response to improve Commons User Experience 

kjgarza:
## Purpose

Improve Commons User Experience of Commons when visiting the individual resources pages (ror.org/, orcid.org/, doi.org/).

addresses:  https://github.com/datacite/lupo/issues/741

## Approach

Caching response of queries that include ID as a query variable. 

Since the metadata and links displayed in Commons of resource tend to change very little in a single period of 24 hrs. And the queries from there are of a single page. We can cache that response and improve the speed of subsequent visits during the same 24hrs window.

## Learning

- Since the metadata and links displayed in Commons of resource tend to change very little in a single period of 24 hrs. And the queries from there are of a single page. We can cache that response and improve the speed of subsequent visits during the same 24hrs window.
- This will also reduce the number of Calls to our Elasticsearch index creating less load in our services overall.

## Types of changes

### DIFF
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class GraphqlController < ApplicationController
+  include Cacheable
   before_action :authenticate_user!
 
   def execute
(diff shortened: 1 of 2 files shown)


app/controllers/graphql_controller.rb
```diff
@@ -11,12 +12,8 @@ def execute
       tracing_enabled: ApolloFederation::Tracing.should_add_traces(headers),
       current_user: current_user,
     }
-    result =
-      LupoSchema.execute(
```
richardhallett, line 15: I was under the impression the apollo server we're using is already caching here.

kjgarza: it is definitely NOT caching. Otherwise, visit the same page twice would be super faster, and it's super slow.

richardhallett (approved):
Alright seems reasonable way to cache on server.

PR #768, opened 2021-12-08
Add certificate filter to potential arguements for repository list in…

jrhoads:
… GraphQL

## Purpose
Previously we could show a list of certificates for facets but could not include them in a query.
This allows for filtering by certificate.

## Approach
Adds arguments to graphql repositories resolver.

## Types of changes


digitaldogsbody:
Looks good to me.

Might be worth fixing the test-on-PR workflow to exclude the now-removed v3 spec (it's already commented out in the merge-to-main workflow) and running the test action again before merge, as there is one further set of tests that would have run after the error made the process bail out. It's only the routing tests, so I don't see any reason why this change would make it would fail, but better to be safe than sorry.

digitaldogsbody (approved):

richardhallett (approved):

jrhoads:
Looks like the tests pass, but there's an issue creating the coverage report. :-(

digitaldogsbody:
> Looks like the tests pass, but there's an issue creating the coverage report. :-(

I don't think the code coverage is run for PRs, just on merge to main.

It looks like the error is still in the v3 tests - I _think_ this is because the workflow on your branch is used in preference to the one in main. Either merging main into your branch, or cherrypicking the change to workflow commit should get things squared up and allow the tests to complete.

PR #770, opened 2022-01-05
Fix a few string array fields on the data_catalog graphql endpoint

jrhoads:
## Purpose
The data_catalog graphql endpoint returned several `#<Hashie::Mash...>` values for string array fields.

## Approach
Unwrap the hashes into string arrays.

## Types of changes


digitaldogsbody (approved):
Looks good to me :)

One question - Do the changes to the GraphQL type file mean the schema needs to be regenerated? Or are things ok since the respective properties already existed and this is just adding a serialization function?

jrhoads:
Not in this case.  The types aren't changing.  The implementation of their response was incomplete

PR #775, opened 2022-01-19
Issue #1431 - Update lupo scripts for managing prefixes/dois.

svogt0511:
## Purpose

closes: datacite/datacite#1431

## Approach

Add rake tasks and code structured similarly to that used for deleting dois, but that changes the doi aasm state from 'findable' to 'registered' instead of deleting the dois.

Modified the 'prefix:delete' task to hide the dois by changing their aasm state from 'findable'  to 'registered' instead of deleting them.  Changed the order of ops in that task to delete the prefix after the dois have changed state.  

Task invocation is still:

```
PREFIX=10.14454 bundle exec rake prefix:delete
```
If we need to delete dois there is a doi:delete_dois_by_prefix task.

- [ ] Is just changing the state sufficient? Are there other things that need to be done when the state is reverted in this way?

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/lib/tasks/prefix.rake
+++ b/lib/tasks/prefix.rake
@@ -111,8 +111,8 @@ namespace :prefix do
     puts "#{count} DOIs with prefix #{ENV['PREFIX']} deleted."
   end
 
-  desc "Delete prefix and hide associated DOIs"
-  task delete_and_hide_dois: :environment do
+  desc "Delete prefix and hide associated DOIs (by changing state from 'findable' to 'registered'"
+  task delete: :environment do
     # These prefixes are used by multiple prefixes and can't be deleted
     prefixes_to_keep = %w(10.4124 10.4225 10.4226 10.4227)
 
--- a/lib/tasks/prefix.rake
+++ b/lib/tasks/prefix.rake
@@ -76,7 +76,7 @@ namespace :prefix do
     puts Prefix.get_registration_agency
   end
 
-  desc "Delete prefix and associated DOIs"
+  desc "Delete prefix and hide associated DOIs (by changing state from 'findable' to 'registered')"
   task delete: :environment do
     # These prefixes are used by multiple prefixes and can't be deleted
     prefixes_to_keep = %w(10.4124 10.4225 10.4226 10.4227)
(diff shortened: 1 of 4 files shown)


lib/tasks/prefix.rake
```diff
@@ -110,4 +110,39 @@ namespace :prefix do
...
+    ClientPrefix.where("prefix_id = ?", prefix.id).destroy_all
+    puts "Client prefix deleted."
+
+    ProviderPrefix.where("prefix_id = ?", prefix.id).destroy_all
+    puts "Provider prefix deleted."
+
+    prefix.destroy
+    puts "Prefix #{ENV['PREFIX']} deleted."
+
+    # hide DOIs
```
digitaldogsbody, line 144: Is it safe to delete the prefix while there are still findable DOIs? Should this operation maybe happen first?

digitaldogsbody (approved):
One non-blocking question about operation order in the "delete and hide" task, but otherwise looks good to me :)

lib/tasks/prefix.rake
svogt0511: I don't know, but it seems safer so I will change the order of operations.  I am also not adding the second task 'delete_and_hide', but just changing the behavior of the delete task and noting the behavior change in the description.  I am thinking that adding another, slightly different method is just overkill and confusing.

jrhoads (approved):
Looks good.

richardhallett (approved):
Looks okay, the deleting of prefix is a bit awkward when the handles still exist but it shouldnt technically matter. Because in theory no updates should need to come in for these DOIs.

svogt0511:
@richardhallett, @digitaldogsbody, @jrhoads - one last question.  I just noticed that running the following script to delete a repository will now cause all of the repository's dois to be put into the 'registered' state.

```
CLIENT_ID=DATACITE.TEST bundle exec rake client:delete
```

Is this the desired outcome?  Or do we really want to delete the dois in this case?

richardhallett:
Sounds desired to me, based on what we discussed at this point we never want to delete DOIs, at least not at this time without some archiving system.

PR #777, opened 2022-01-20
Issue #1435 - Lupo handling of include_deleted flag on export requests…

svogt0511:
…s (salesforce).

## Purpose

closes: datacite/datacite#1435

## Approach

Allow API export requests to pass the include_deleted flag through to elastic search.  This is in support of the salesforce daily sync which uses these requests to export providers, repositories, and contacts from the DataCite database to the Salesforce database.

None

## Types of changes

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -648,28 +648,16 @@ class Client < ApplicationRecord
       "deleted_at" => deleted_at ? deleted_at.iso8601 : nil,
     }
 
-    Rails.logger.info "SKV - TO_JSONAPI BEGIN: uid, doi_counts, dois_total, dois_current_year, dois_last_year"
-    Rails.logger.info uid
-    Rails.logger.info doi_counts
-    Rails.logger.info dois_total
-    Rails.logger.info dois_current_year
-    Rails.logger.info dois_last_year
-    Rails.logger.info "SKV - TO_JSONAPI MIDDLE: attributes"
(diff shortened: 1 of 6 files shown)


app/models/client.rb
```diff
@@ -648,28 +648,16 @@ def to_jsonapi
...
-    Rails.logger.info attributes
-    Rails.logger.info "SKV - TO_JSONAPI END"
-
     { "id" => symbol.downcase, "type" => "clients", "attributes" => attributes }
   end
 
-  def self.export(query: nil)
-    Rails.logger.info "SKV - EXPORTING CLIENTS - BEGIN"
-
+  def self.export(query: nil, include_deleted: false)
```
richardhallett, line 654: This now changes the default behaviour, it was originally to always include deleted, which was what was generally preferred for the export. Is the intention now to make it exclude and only include optionally?

svogt0511: Yes, it does change the default behavior.  The intention was to make its behavior consistent with the rest of the API and then change where it is used, which is hopefully only in the Salesforce sync job. If you leave the default behavior, including deleted, then there is no way to affect the behavior with this option.  Currently the option's value has no affect.  I can change it so it pays attention to include_deleted=true and include_deleted=false.

Do we have this documented anywhere?

svogt0511:
Closing the pull request until we resolve issues around it.  Keeping the branch around until then.

PR #784, opened 2022-02-15
Re-enable tests

richardhallett:
To renable tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,14 +57,14 @@ jobs:
             ${{ runner.os }}-gems-
       - name: Install
         env:
-          MYSQL_PORT: ${{ job.services.mysql.ports[3306] }} 
+          MYSQL_PORT: ${{ job.services.mysql.ports[3306] }}
         run: |
           cp .env.build .env
           gem install bundler
           bundle config path vendor/bundle
           bundle install --jobs 4 --retry 3
           sudo /etc/init.d/mysql start
-          # bundle exec rails db:setup RAILS_ENV=test      
+          bundle exec rails db:setup RAILS_ENV=test
       - name: Lint and Test
         env:
           MYSQL_PORT: ${{ job.services.mysql.ports[3306] }}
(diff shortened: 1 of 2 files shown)


svogt0511 (approved):
Looks good to me.  Just one question: is the passenger_max_request_queue_size config meant to be a part of this mod?

.github/workflows/release.yml
```diff
@@ -73,7 +73,7 @@ jobs:
           MEMCACHE_SERVERS: "localhost:11211"
         run: |
           bundle exec rubocop
-          # bundle exec rspec
+          bundle exec rspec
```
jrhoads, line 76: On the deploy workflow, instead of running just rspec it has a series of specific specfiles it executes

https://github.com/datacite/lupo/blob/9f4380dfbc19515e040771d373c33fdc88331519/.github/workflows/deploy.yml#L80

Should we do the same here?

svogt0511: I did that a while ago.  I thought I was seeing flakiness in the test run due to resource limitations so I split up the run.  It may run ok in the test environment.  Give it a try.  Probably better to have just the one command.

digitaldogsbody (approved):
> Looks good to me. Just one question: is the passenger_max_request_queue_size config meant to be a part of this mod?

Seconding this. I think we need to have a think about how we handle surges to the API - the increased queue size might not work very well for Lupo, where requests are a lot more complex (and therefore potentially slower) than the Content Negotiation service where we previously used this mitigation, so it might be worth removing this before merge, pending more discussion.

.github/workflows/release.yml
digitaldogsbody: I think it should be ok to roll this back into one command - that's certainly how I usually run it locally and I haven't seen any issues.

richardhallett:
This is what happens when I do a commit while talking on a call. No that should not be in this. I must have been playing with it and forgot.
I agree that we do not want to increase this without further thought/testing.

jrhoads (approved):

PR #786, opened 2022-02-23
Remove password from jwt.

svogt0511:
## Purpose

closes: none.

Fix potential jwt issue.  Just deleted the addition of the password and fixed tests.  

I was able to test handle creation locally.  Looks like it worked.  We can also test that in stage.

## Types of changes


jrhoads (approved):
Looks good

kjgarza:
I think LGTM. But to be sure and safe in advance,  why don't you just create a JWT, using the [JWT generate cli](https://datacite.atlassian.net/wiki/spaces/DAT/pages/905609220/How+to+generate+a+jwt+for+users) (I think it uses the exact same method) and then create a DOI in `production` under **datacite.blog** or **datacite.datacite** , that way you can be totally sure it will work.

svogt0511:
@kjgarza Will do that, compare them and let you know.

digitaldogsbody (changes requested):
There might be a couple of other changes to make in authenticatable.rb sure this is completely eliminated:

L314(old) / L311(new): Either remove the password attribute or make it always nil
L337(old) / L334(new): Remove password here as well? How is the ALB token used?
L364-L368(old) / L361-L365(new): Refactor the merge in the same way as the existing change made by the PR in case there are any instances of the class method being called anywhere rather than just on an instantiated member of the class.

svogt0511:
@digitaldogsbody Making the changes and doing additional testing.

digitaldogsbody (approved):

svogt0511:
@kjgarza @digitaldogsbody @jrhoads - Changes made.  Passes the test suite.  Created a token and then using that token, created a doi in datacite.datacite production.  Handle resolves just fine.  This is it.  Here goes.  Thanks for looking it over.

Issue #601, opened 2020-08-12
[Graphql] CitationsOverTime gives weird counts distribution


kjgarza:
### Expected Behaviour

CitationsOverTime should return a distribution of citations binned by the publication date of the newest doi in the relation of the events.

This was implemented using `citation_year` in the index:

```
 def citation_year
    "" unless INCLUDED_RELATION_TYPES.include?(relation_type_id)
    subj_publication = subj['date_published'] || (date_published(subj_id) || year_month)
    obj_publication =  obj['date_published']  || (date_published(obj_id) || year_month)
    [subj_publication[0..3].to_i, obj_publication[0..3].to_i].max
  end


```

We discussed this feature in:

- https://github.com/datacite/lupo/issues/313
- https://github.com/datacite/lupo/pull/330
- https://github.com/datacite/lupo/commit/2390b78404792d6aba354f22fff923c930e3dd1b

### Current Behaviour

CitationsOverTime is retuning  a distribution of citations binned by the `occurred_at` date in the event.

see: https://github.com/datacite/lupo/blob/a2a6743/app/models/doi.rb#L1171

```
  # remove duplicate citing source dois,
  # then show distribution by year
  def citations_over_time
    citation_events.pluck(:occurred_at, :source_doi).uniq { |v| v[1] }
      .group_by { |v| v[0].utc.iso8601[0..3] }
      .map { |k, v| { "year" => k, "total" => v.length } }
      .sort_by { |h| h["year"] }
  end


```

`occurred_at` is updated by the EventData agent and thus cannot be used to bin results.

### Steps to Reproduce

query for 

```
{
  work(id: "10.3334/cdiac/00001_v2015"){
    publicationYear
    citationCount
    citationsOverTime{
      year
      total
    }
    citations{
			published{
        id
        count
      }
    }
        references{
published{
        id
        count
      }
    }
  }
}
```

### Context (Environment)

Citation_year was implemented back in august 2019. look at 

- https://github.com/datacite/lupo/issues/313
- https://github.com/datacite/lupo/pull/330
- https://github.com/datacite/lupo/commit/2390b78404792d6aba354f22fff923c930e3dd1b

### Possible Implementation

aggregate citations using `citation_year`

mfenner:
Please keep in mind that we made significant changes to the Event/DOI integration in January 2020, including adding a `citationsOverTime` property. I suggest to make the following changes:

[...]

Issue #150, opened 2018-12-01
The attribute 'lang' is not allowed


mfenner:
<img width="991" alt="bildschirmfoto 2018-12-01 um 09 43 08" src="https://user-images.githubusercontent.com/23151/49326120-9243bd80-f54d-11e8-9d1d-42d70f41a349.png">

mfenner:
This is the expected behviour according to the DataCite Metadata Schema documentation. The attribute should be `xml:lang`.

<img width="1142" alt="bildschirmfoto 2018-12-01 um 10 12 02" src="https://user-images.githubusercontent.com/23151/49326381-b0132180-f551-11e8-93f6-84e81ce5fa33.png">

PR #790, opened 2022-03-07
Doi estimate field

svogt0511:
## Purpose

closes: none

Add doi_estimate field to the database and include it in API requests.

Related to:

- Bracco [pull request #601](https://github.com/datacite/bracco/pull/601)
- Mastino [pull request #150](https://github.com/datacite/mastino/pull/150)

This mod is applied first.  Then the bracco mod (with or without the feature flag). Then the mastino (salesforce) mod can be applied.

Run the DB migration.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -150,7 +150,7 @@ class Provider < ApplicationRecord
   # validates :voting_contact, contact: true
   # validates :billing_information, billing_information: true
 
-  validates :doi_estimate_year_one, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :doi_estimate_year_one, numericality: { only_integer: true, greater_than_or_equal_to: 0 } if :member_type === "consortium_organization"
 
   strip_attributes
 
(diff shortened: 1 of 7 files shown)


app/models/provider.rb
```diff
@@ -912,6 +947,11 @@ def set_defaults
       self.billing_information = {} if billing_information.blank?
       self.consortium_id = nil unless member_type == "consortium_organization"
       self.non_profit_status = "non-profit" if non_profit_status.blank?
+      if member_type === "consortium_organization"
```
jrhoads, line 950: Is this supposed to be a `==` or are you explicitly checking to see if it is included in a list?

https://dev.to/baweaver/understanding-ruby-triple-equals-2p9c

spec/models/provider_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+require "pp"
```
jrhoads, line 4: Maybe leftover from debugging?

app/models/provider.rb
svogt0511: Good catch.  Will fix that.

spec/models/provider_spec.rb
svogt0511: Yup. Will take care of that, as well.

digitaldogsbody (approved):
Looks good to me :)

PR #791, opened 2022-03-09
Combined index for Client and Re3Data Repositories

jrhoads:
## Purpose
Create a combined model and elastic-search index for Client and Re3Data Repositories

closes:
 - datacite/datacite#1395
 - datacite/datacite#1393

## Approach
Create a model that effectively joins client_ids and re3data_dois

Use elastic search to index this model with fields from both repository types

When fields are present in both models (name, description, etc) prefer Client Repository information

Create a unique identifier for each join.
- [ ] What form should the unique identifier take?
   - Relatively small total count (~5000)
   - Currently using [Hashids](https://hashids.org/ruby/) on the internal identifier
   - Also considered [NanoIDs](https://zelark.github.io/nano-id-cc/) 

## Learning
These two videos have been a great help
- [ElasticSearch and Ruby on Rails - Part 1](https://www.youtube.com/watch?v=6i62s2v_2Og)
- [ElasticSearch and Ruby on Rails - Part 2](https://www.youtube.com/watch?v=hHoDKtN8kiI)

## Types of changes

### DIFF
--- a/app/models/reference_repository.rb
+++ b/app/models/reference_repository.rb
@@ -1,0 +1,11 @@
+class ReferenceRepository < ApplicationRecord
+    include Hashid::Rails
+    validates_uniqueness_of :re3doi, :allow_nil => true
+
+    def client
+        @client ||= Client.find_by_id(self[:client_id])
+    end
+    def re3repo
+        @re3repo ||= DataCatalog.find_by_id(self[:re3doi])
+    end
+end
--- a/app/models/reference_repository.rb
+++ b/app/models/reference_repository.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ReferenceRepository < ApplicationRecord
+  include Indexable
   include Elasticsearch::Model
   include Elasticsearch::Model::Callbacks
   include Hashid::Rails
--- a/app/models/reference_repository.rb
+++ b/app/models/reference_repository.rb
@@ -35,6 +35,17 @@ class ReferenceRepository < ApplicationRecord
             indexes :description
             indexes :pid_system, type: :keyword
             indexes :url
+            indexes :keyword, type: :keyword
+            indexes :subject
+            indexes :contact
+            indexes :language, type: :keyword
+            indexes :certificate, type: :keyword
+            indexes :data_access, type: :keyword
+            indexes :data_upload, type: :keyword
+            indexes :provider_type, type: :keyword
+            indexes :repository_type, type: :keyword
+            indexes :data_upload_licenses, type: :keyword
+            indexes :software, type: :keyword
         end
     end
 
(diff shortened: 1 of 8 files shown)


svogt0511:
If this is pushed to production, please make sure to run the database migration.  My pull request #790 is merged to master but not yet in production and requires a DB migration when it is pushed to production.  Thanks.

svogt0511:
I love the research on the uuids.

lib/tasks/repository.rake
```diff
@@ -0,0 +1,46 @@
...
+      progressbar.increment
+      ReferenceRepository.find_or_create_by(
+        client_id: c.symbol,
+        re3doi: c.re3data_id
+      )
+    end
+  end
+
+  desc "Load all Re3data Repositories into Reference Repostories"
+  task :load_re3data_repos, [:pages] => :environment do |t, args|
```
richardhallett, line 22: I'd move the actual fetching from re3data and such into the model, then this rake task calls that.
That way you can open the import of re3data (or datacite client) code to a background job.
i.e. for say updating based on when one of our clients gets updated.

app/models/reference_repository_denormalizer.rb
```diff
@@ -0,0 +1,129 @@
...
+  attr_reader :repository
+
+  def initialize(repository)
+    @repository = repository
+  end
+
+  def doi_as_url
+    doi = @repository.re3doi
+    return nil if doi.blank?
+    "https://doi.org/#{doi.downcase}"
```
richardhallett, line 13: These are always live DOIs regardless of system right? re3data has no staging like we do (handle.stage.datacite.org)

app/models/reference_repository.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+class ReferenceRepository < ApplicationRecord
```
richardhallett, line 3: You may wish to bring in the Indexable Concern.
It does bring a lot with it so I appreciate it's fairly complex, but it will give you the stuff regarding switching back and front indexing (if desired)
And also has some helper functions for searching / aggregation though granted it is a bit mixed up between the various classes (something that i'd like to refactor one day)

richardhallett (approved):
As a general approach I think this makes sense.
Using the HashID function based on the database primary key (i.e. the integer value) is fine I think for this table as it's unlikely to change and this therefore makes it pretty consistent.

I'm happy to approve as it stands though maybe needs a bit of tweaking around some of import functions and such.

app/models/reference_repository_denormalizer.rb
jrhoads: Correct

app/models/reference_repository.rb
jrhoads: I brought it in.
Works well

lib/tasks/repository.rake
jrhoads: This is for initial load much like a db:seed.  All of the indexing and updating for individuals can be done bone in background jobs.

PR #803, opened 2022-04-20
ReferenceRepository graphql

jrhoads:
## Purpose
Add functionality to the GraphQL endpoint to allow searching ReferenceRepositories (combined client and re3Data repositories)

closes: 
datacite/datacite#1395
datacite/datacite#1394
datacite/datacite#1393

## Types of changes


app/graphql/schema.graphql
```diff
@@ -8418,6 +8420,166 @@ type Query {
...
+  Description of the repository
+  """
+  description: String
+
+  """
+  The number of downloads according to the Counter Code of Practice.
+  """
+  downloadCount: Int
+
+  """
```
digitaldogsbody, line 8472: Minor typo - language

app/graphql/schema.graphql
```diff
@@ -8418,6 +8420,166 @@ type Query {
...
+  The repository type(s)
+  """
+  repositoryType: [String!]
+
+  """
+  The name of the software that is used to run the repository
+  """
+  software: [String!]
```
digitaldogsbody, line 8510: The r3d schema has a `minOccurs` of 0 for this property - should it definitely be non-null?

app/graphql/schema.graphql
```diff
@@ -8418,6 +8420,166 @@ type Query {
...
+type ReferenceRepository {
+  """
+  Repository alternate names
+  """
+  alternateName: [String!]
+
+  """
+  The certificate(s) for the repository
+  """
+  certificate: [String!]
```
digitaldogsbody, line 8435: The r3d schema has a `minOccurs` of 0 for this property - should it definitely be non-null?

app/graphql/schema.graphql
```diff
@@ -8418,6 +8420,166 @@ type Query {
...
+
+  """
+  The unique identifier for the client repository
+  """
+  clientId: ID
+
+  """
+  Repository contact information
+  """
+  contact: [String!]
```
digitaldogsbody, line 8450: The r3d schema has a `minOccurs` of 0 for this property - should it definitely be non-null?

app/graphql/schema.graphql
```diff
@@ -8418,6 +8420,166 @@ type Query {
...
+
+  """
+  Repository name
+  """
+  name: String!
+
+  """
+  PID Systems
+  """
+  pidSystem: [String!]
```
digitaldogsbody, line 8485: The r3d schema has a `minOccurs` of 0 for this property - should it definitely be non-null?

app/models/reference_repository.rb
```diff
@@ -9,6 +9,15 @@ class ReferenceRepository < ApplicationRecord
...
 
   validates_uniqueness_of :re3doi, allow_nil: true
+  #
+  # use different index for testing
+  if Rails.env.test?
+    index_name "refernce_repositories-test"
+  elsif ENV["ES_PREFIX"].present?
+    index_name "refernce_repositories-#{ENV['ES_PREFIX']}"
+  else
+    index_name "refernce_repositories"
```
digitaldogsbody, line 19: "reference" typo?

digitaldogsbody (changes requested):
The non-null fields in the GraphQL schema are a bit odd, because the reference_repository_type file explicitly defines those fields as nullable, so I am not sure why the schema generator is appending the ! non-nullable operator to them - my comments are because the schema was the first file shown in the review tab, so I hadn't clocked that it was different in the type file. 

Might be worth checking out why that happened anyway in case the non-nullables in the schema have any implications on the query function? I would not say this is a merge-blocking change, but my experience with GraphQL is limited.

I think that the typo in the app/models/reference_repository.rb file _should_ be fixed before merge, lest we accidentally create an extra ES index (or throw a bunch of errors because the index can't be found)

Other than the ES index issue, it looks good to me :slightly_smiling_face:

app/graphql/types/reference_repository_type.rb
```diff
@@ -0,0 +1,206 @@
+
+# frozen_string_literal: true
+
+class ReferenceRepositoryType < BaseObject
+  description "Information about repositories"
```
kjgarza, line 5: I was not anticipating a new PUBLIC interface (ReferenceRepositoryType) for this. particularly when it seems, in a first view, exactly the same fucntionality exposed in RepositoryType. I was anticiapating you will just replace it. 

Particularly with public interfaces I think we should avoid "exposing complexity" of the our intricated system. Could we discuss this @jrhoads?

kjgarza: Basically, if the GraphpQL user (we are not the only user) needs Repositories they go to a single place, and if they need a subset they filter using the same type.

jrhoads: Happy to discuss (and change).  I had some concerns about someone wanting to interact with a "Client-type" repository from something like Fabrica or the like.  I'm not sure if that is a realistic concern or not.

app/models/reference_repository.rb
jrhoads: Good catch.

app/models/reference_repository.rb
```diff
@@ -9,6 +9,15 @@ class ReferenceRepository < ApplicationRecord
   before_save :force_index
 
   validates_uniqueness_of :re3doi, allow_nil: true
+  #
```
kjgarza, line 12: I'm not sure whether this is part of this API but, how is one ReferenceRepository instace created when someone create a Client/Repo in Fabrica/REST-API?   

I see ReferenceRepository.find_or_create_by is triggered in rake tasks and in rspec but when running in production how are the new repositories added to the ReferenceRepository index/db? 

Sorry If I'm missing something.

app/models/reference_repository.rb
jrhoads: Fixed

app/models/reference_repository.rb
jrhoads: Fair point.  You are not missing anything.  It does not currently auto-update or auto-add on save.

Even though this PR adds a lot, I was planning on making sure the bones were in place before refining the creation/update steps an a future PR.  If that sounds OK, I can add a GH issue to make sure that intention is recorded.

app/graphql/schema.graphql
digitaldogsbody: Aha! That makes a lot of sense. I had missed the difference between `[String!]` and `[String]!`

digitaldogsbody: and, indeed, `[String!]!`

digitaldogsbody (approved):

app/models/reference_repository.rb
kjgarza: well I guess it's just a callback on create like in:

https://github.com/datacite/lupo/blob/449db11c9eec14573b80a6fd7918c74dee02af6e/app/models/doi.rb#L129

jrhoads: In our zoom you showed that it wouldn't be a large lift to add this in.  Will do and update this PR.

app/graphql/types/reference_repository_type.rb
jrhoads: We had a conversation on zoom and I will update the GraphQL endpoints so this replaces the Repository endpoint rather than adding a new ReferenceRepository endpoint.

app/graphql/types/repository_connection_with_total_type.rb
```diff
@@ -5,12 +5,10 @@ class RepositoryConnectionWithTotalType < BaseConnection
   field_class GraphQL::Cache::Field
 
   field :total_count, Integer, null: true, cache: true
-  field :years, [FacetType], null: true, cache: true
-  field :members, [FacetType], null: true, cache: true
```
kjgarza, line 9: Do we need to remove the Member facet functionality from this interface? I mean this information does still exist for all the repos that have datacite dois.

app/graphql/types/repository_type.rb
```diff
@@ -54,131 +92,13 @@ class RepositoryType < BaseObject
           "The number of downloads according to the Counter Code of Practice."
   field :citation_count,
         Integer,
-        null: true, description: "The number of citations."
-
-  field :datasets,
```
kjgarza, line 59: I'm not sure, why you need to delete these ones for the repository type. I know the repository page in commons won't use it. But why would you remove it from the API completely?

kjgarza: i guess the same comment goes for software, publications, dmps, etc.

app/graphql/types/repository_connection_with_total_type.rb
jrhoads: implemented

app/graphql/types/repository_type.rb
jrhoads: Reverted

richardhallett (approved):

PR #804, opened 2022-04-22
Use non-depricated way to specify the path for bundle installs

jrhoads:
## Purpose
Build was failing due to permissions issues.

## Approach
Add line for `bundle config set --local path 'vendor/bundle'`

Specifically install the version of bundler that matches the Gemfile.lock
 
## Learning
Turns out that `gem update --system` was updating the version of bundler. So when we bundle installed, it discovered that the version in the Gemfile.lock was out of sync with system bundler and "helpfully" tried to install the version of bundler in the Gemfile.lock.  However, the user was set to `app` which did not have permissions to install to system resources.  Bundler then restarted and tried to install anyway using the most recent bundler.

Additionally since we had been using the deprecated flag to set the local path, when all these restarts happened, that flag got lost.  Fortunately there was a message with the deprecation that indicated that we should use the new `bundle config` version fo the command.

Additionally we reinstated the installation of the specific version of bundler that is in lock step with the gemfile.  We recommend that folks feel free to use up to date versions of bundler to generate the Gemfile.lock but that they also update the versions of bundler in the docker file. 

## Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,9 @@ RUN mkdir -p vendor/bundle && \
     chown -R app:app . && \
     chmod -R 755 . && \
     gem update --system && \
-    /sbin/setuser app bundle install --path vendor/bundle
+    gem install bundler:2.3.10 && \
+    /sbin/setuser app bundle config set --local path 'vendor/bundle' && \
+    /sbin/setuser app bundle install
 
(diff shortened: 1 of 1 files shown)


svogt0511 (approved):
Looks good to me.

PR #809, opened 2022-05-12
GraphQL returns nil for creators/contributors properties with hash for affiliation 

### LINKED ISSUE #802: GraphQL returns null for "creators" property when a creator in the DOI contains schema v3 affiliation metadata
##Expected Behaviour
GraphQL responses contain creator metadata where present. 

##Current Behaviour
GraphQL responses return a `null` value in the creators property for DOIs that contain schema v3 affiliation metadata in a creator property. 

Example GraphQL response for DOI 10.7914/sn/1a_2014:

`20: {id: "https://doi.org/10.7914/sn/1a_2014", doi: "10.7914/sn/1a_2014", identifiers: [],…}
citationCount: 0
claims: []
container: {identifier: null, identifierType: null, title: null, __typename: "Container"}
**creators: null**
descriptions: [{,…}]
doi: "10.7914/sn/1a_2014"
downloadCount: 0
fieldsOfScience: []
id: "https://doi.org/10.7914/sn/1a_2014"
identifiers: []
language: {id: null, name: null, __typename: "Language"}
publicationYear: 2014
publisher: "International Federation of Digital Seismograph Networks"
registered: "2015-01-28T00:35:37Z"
registrationAgency: {id: "datacite", name: "DataCite", __typename: "RegistrationAgency"}
repository: {id: "iris.iris", name: "Incorporated Research Institutions for Seismology", __typename: "Repository"}
rights: []
titles: [{title: "Mining-induced seismicity network at mine Prosper-Haniel, Bottrop", __typename: "Title"}]
types: {resourceTypeGeneral: "Other", resourceType: "Seismic Network", __typename: "ResourceType"}
version: null
viewCount: 0
__typename: "Work"`

`"creators": [
{
"name": "Kasper D. Fischer",
"nameType": "Personal",
**"affiliation": [
"Ruhr-University Bochum, Germany"
],**
"nameIdentifiers": []
},
{
"name": "Dennis Wlecklik",
"nameType": "Personal",
"affiliation": [
"Ruhr-University Bochum, Germany"
],
"nameIdentifiers": []
},
{
"name": "Sebastian Wehling-Benatelli",
"nameType": "Personal",
**"affiliation": [
"Ruhr-University Bochum, Germany"
],**
"nameIdentifiers": []
},
{
"name": "Wolfgang Friederich",
"nameType": "Personal",
**"affiliation": [
"Ruhr-University Bochum, Germany"
],**
"nameIdentifiers": []
}
],`

Also see DOI 10.7914/sn/xc_2015

##Steps to Reproduce
See above; see multiple errors in this GraphQL search response: https://commons.datacite.org/doi.org?query=types.resourceType%3ASeism*

##Context (Environment)
Commons production, lupo production.

##Screenshots

##Further details
Because the GraphQL response includes an error, none of the results are rendered in Commons. See example search: https://commons.datacite.org/doi.org?query=types.resourceType%3ASeism*

#Proposal
##Hypothesis
The issue can likely be derived from changes in the GraphQL response construction that do not account for schema v3 differences. In this case, the GraphQL schema expects affiliation properties with id and name properties but receives a string. 

See this line in the GraphQL schema: https://github.com/datacite/lupo/blob/2c8a5799a470be53891a6510d0fd65d638f85c79/app/graphql/schema.graphql#L98

See this line in the GraphQL mapping for creators: https://github.com/datacite/lupo/blob/2c8a5799a470be53891a6510d0fd65d638f85c79/app/graphql/types/doi_item.rb#L424

##Possible Implementation

codycooperross:
Fixes #802

## Purpose
Affiliation properties stored as hashes caused nil creators types in GraphQL responses. Downstream, this caused some searches in Commons to fail. 

closes: #802 

## Approach
Affiliation properties within creators and contributors are now wrapped in an array in the doi_item type. 

## Types of changes


app/graphql/types/doi_item.rb
```diff
@@ -460,9 +460,9 @@ def contributors(**args)
         "given_name" => c.fetch("givenName", nil),
         "family_name" => c.fetch("familyName", nil),
         "affiliation" =>
-          c.fetch("affiliation", []).map do |a|
+          Array.wrap(c.fetch("affiliation", [])).map do |a|
             { "id" => a["affiliationIdentifier"], "name" => a["name"] }.compact
-          end,
+          end.compact,
```
digitaldogsbody, line 465: Will this `.compact` fire properly in all the situations where there might be a `null` value in the array?

If for some reason there is a `null` in the wrapped array, this looks like it would currently pass through to the `map` function first, which might lead to an error, rather than being removed by `.compact`.

I think the way to prevent that is to move (or add another) `.compact` between the `Array.wrap()` and `map`, but the big caveat here is that I am still learning about the fun intricacies of Ruby in this regard, so it would probably be worth writing a spec test first to see if it would indeed fail under the circumstances above before making any changes,

app/graphql/types/doi_item.rb
```diff
@@ -434,9 +434,9 @@ def creators(**args)
...
```
digitaldogsbody, line 439: Same question here as for the block in `contributors`

jrhoads (approved):

digitaldogsbody (approved):
The fix looks good to me. I have one question about the functionality when there is an unexpected `null` in the data, but it's not blocking for merge :)

app/graphql/types/doi_item.rb
richardhallett: I imagine actually it'll be fine. It's a little odd but I imagine
[nil] maps to { "id" => nil, "name" => nil } which compact sto just [nil] again.
So the last compact then makes it []

The other way to solve this would be to just compact before the map I think? I'd have to actually run the code but I'd have thought if you compact before doing map then it wouldnt even go into map as it'd be nil.

Although with what Cody has done this does potentially catch scenario where you have a hash which has nil values for it's parts.
so:
[{ "affiliationidentifier" => nil, "name" => nil}] maps to { "id" => nil, "name" => nil } which compacts to nil again.

So in summary maybe go with this, for extra care like Mike said you could write a spec to handle these scenarios.

richardhallett (approved):

app/graphql/types/doi_item.rb
codycooperross: For context, the general strategy here was to mimic the serialization on the REST side: https://github.com/datacite/lupo/blob/344ccdecd1998518411084d75140c56e78ea3a6d/app/serializers/datacite_doi_serializer.rb#L127 

Testing for nil values within an array in the affiliation property is a little challenging because I don't think the XML generation code can handle the nil values. I get this error when trying:

```     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./app/models/concerns/crosscitable.rb:134:in `update_xml'
     # ./spec/graphql/types/work_type_spec.rb:897:in `block (3 levels) in <top (required)>'```

I'm sure there's a workaround — can explore more if that makes sense.

PR #824, opened 2022-06-08
Fix Wrong relationtypes

kjgarza:
## Purpose

setting types according to: https://support.datacite.org/docs/relationtype_for_citation

closes: https://github.com/datacite/datacite/issues/1413

- [ ] A separated task would be to clean the events wrongly computed. 

## Learning

- [ ] The arrays in the code need better naming.

- [ ] 2Million events might be affected (0.014%)

image

## Types of changes

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -86,10 +86,10 @@ class Event < ApplicationRecord
   ].freeze
 
   # renamed to make it clearer that these relation types are grouped together as references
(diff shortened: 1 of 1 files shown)


digitaldogsbody (approved):
We need to have a think about how we will repair existing incorrect relationships. One thing that came to light with the Cites/IsCitedBy is that just looking at records that have changed since the array change was introduced didn't necessarily catch all of the historic events.

PR #826, opened 2022-06-10
Updating client trigger reference_repository update.  Respects uniqueness of re3dois

jrhoads:
Add tests for reference_repository creation/updates/delete

## Approach
On `update_from_client` checks for validation (uniqueness of re3dois)
If this fails, then delete the duplicate reference_repository
Extra check to make sure that duplicate does not already have a client_id associated with it.

## Types of changes

### DIFF
--- a/app/models/reference_repository.rb
+++ b/app/models/reference_repository.rb
@@ -52,6 +52,9 @@ class ReferenceRepository < ApplicationRecord
     rr = ReferenceRepository.find_or_create_by(client_id: client.uid)
     if client.re3data_id && (not rr.re3doi)
       rr.re3doi = client.re3data_id
+      if !rr.validate
+        ReferenceRepository.find_by(re3doi: client.re3data_id, client_id: nil).try(:destroy)
+      end
       rr.save
     else
       rr.touch
--- a/spec/models/reference_repository_spec.rb
+++ b/spec/models/reference_repository_spec.rb
@@ -2,6 +2,96 @@
 
 require "rails_helper"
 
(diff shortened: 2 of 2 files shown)


app/models/reference_repository.rb
```diff
@@ -52,6 +52,9 @@ def self.update_from_client(client)
     rr = ReferenceRepository.find_or_create_by(client_id: client.uid)
     if client.re3data_id && (not rr.re3doi)
       rr.re3doi = client.re3data_id
+      if !rr.validate
+        ReferenceRepository.find_by(re3doi: client.re3data_id, client_id: nil).try(:destroy)
```
kjgarza, line 56: you might not need the if-clause if you put this line at the very beginning of the method. I think [.try method](https://apidock.com/rails/v3.2.1/Object/try) will just return a Nill if the object doesn't exist. And delete it if it exists.

app/models/reference_repository.rb
```diff
@@ -52,6 +52,9 @@ def self.update_from_client(client)
     rr = ReferenceRepository.find_or_create_by(client_id: client.uid)
```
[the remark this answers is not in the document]
kjgarza: actually no. i think you won't need this

app/models/reference_repository.rb
jrhoads: Thanks for taking a look at this.

I think that you are right.  The code should technically work without clause, but it does help explain why we are eventually running the destroy method.

I think will help "future us" understand this code when we look back on it in the future.

PR #832, opened 2022-06-27
Addresses error where some author aggs with multiple nameIdentifiers result in array index error

codycooperross:
## Purpose

Some Commons searches and pages would report an error page due to an issue where ["works", "authors"] was nil. 

closes: 1418 1309 1359

## Approach

ElasticSearch GQL searches aggregate authors by creators.nameIdentifiers.nameIdentifier, like so: https://github.com/datacite/lupo/blob/bab0c144fcc68489ec8ba6d2287562a3713734cd/app/models/doi.rb#L614

This can result in multiple nameIdentifiers represented in the author aggs for the same person, some of which are not necessarily ORCIDs. These are represented like so:

`{
                                    "name": "Gomeseria, Ronald",
                                    "nameIdentifiers": [
                                        {
                                            "nameIdentifier": "https://osf.io/8kzbu/"
                                        },
                                        {
                                            "nameIdentifier": "https://orcid.org/0000-0002-3776-4755"
                                        }
                                    ]
                                }`

The current code attempts to find the relevant creator by matching the agg key to the creators list. This select would sometimes break before finding the correct id, leaving a blank matched_creator variable and thus an index error at this line:
https://github.com/datacite/lupo/blob/e9d55ad379c568ad55792833a67b23c187bfa960/app/graphql/types/base_connection.rb#L217

The pull request attempts to: 1) rewrite the select behavior to accommodate situations where multiple nameIdentifiers exist for a given creator, 2) filter out identifiers that are not ORCIDs per the tooltip explanation of the facet functionality in Commons, and 3) make the facet_by_authors function a bit safer. 

It's unclear if this line is a "best practice" approach to filtering ORCIDs: https://github.com/datacite/lupo/blob/8f309b83c225156e6f9d12c2b744e467afe49a1f/app/controllers/concerns/facetable.rb#L412

Borrows regex from here:
https://github.com/datacite/lupo/blob/dd116fb95ce417ff53fc05773d509333aa34ab6d/app/graphql/types/base_connection.rb#L66

## Types of changes

### DIFF
--- a/app/controllers/concerns/facetable.rb
+++ b/app/controllers/concerns/facetable.rb
@@ -411,6 +411,10 @@ module Facetable
       arr.map { |hsh|
         orcid_id = %r{\A(?:(http|https)://(orcid.org)/)(.+)\z}.match?(hsh["key"]) && hsh["key"]
 
+        if orcid_id.nil?
+          next
+        end
+
         # The aggregation query should only return 1 hit, so hence the index
         # into first element
         creators = hsh.dig("authors", "hits", "hits")[0].dig("_source", "creators")
--- a/app/controllers/concerns/facetable.rb
+++ b/app/controllers/concerns/facetable.rb
@@ -409,8 +409,8 @@ module Facetable
 
     def facet_by_authors(arr)
       arr.map { |hsh|
-        orcid_id = hsh["key"]
-        
+        orcid_id = %r{\A(?:(http|https)://(orcid.org)/)(.+)\z}.match?(hsh["key"]) && hsh["key"]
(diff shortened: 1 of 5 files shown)


jrhoads:
I believe this work would only affect the REST api.

Commons runs on the GraphQL api. 
There is an analogous function by the same name `facet_by_authors` in `app/graphql/types/base_connection.py` that is used in the GraphQL api.

jrhoads (approved):
Great job.  I anticipate this will solve many errors we've been seeing in Commons.

Issue #828, opened 2022-06-10
Page-based pagination returns unusable links once you hit 10k records


digitaldogsbody:
The `page[number]=X&page[size]=Y` method of querying has a limit of 10k records due to ElasticSearch. This is not an issue, as you can use cursor-based pagination for larger result sets.

However, the page based pagination will include links for pages beyond the limit, which are obviously inaccessible. To compound matters, the last page of results is returned again, with a `next` link again pointing beyond the accessible end of the result set. This could easily result in an infinite loop if a user is expecting (quite sensibly) that `next` will be empty at the end of the results.

Recreate:

1. Query [https://api.datacite.org/dois?page[size]=100&page[number]=1&query=doi:10.18710/B7BWXO/*](https://api.datacite.org/dois?page[size]=100&page[number]=1&query=doi:10.18710/B7BWXO/*)
2. The `meta` section correctly contains a `totalPages` of 10
```json
  "total": 27723,
  "totalPages": 10,
  "page": 1,
```
3. Request page 10: [https://api.datacite.org/dois?page[size]=100&page[number]=10&query=doi:10.18710/B7BWXO/*](https://api.datacite.org/dois?page[size]=100&page[number]=10&query=doi:10.18710/B7BWXO/*)
4. The `links` section contains (`next` url decoded for ease of understanding):
```json
  "self": "https://api.datacite.org/dois?page[size]=1000&page[number]=10&query=doi:10.18710/B7BWXO/*",
  "next": "https://api.datacite.org/dois?page[number]=11&page[size]=1000&query=doi:10.18710/B7BWXO/*"
```
5. Attempt to retrieve the `next` link
6. Page 10 is returned again and the `next` link is again pointing to the non-existent Page 11 (although `self` has updated):
```json
  "total": 27723,
  "totalPages": 10,
  "page": 10,
```
```json
  "self": "https://api.datacite.org/dois?page[size]=1000&page[number]=11&query=doi:10.18710/B7BWXO/*",
  "next": "https://api.datacite.org/dois?page[number]=11&page[size]=1000&query=doi:10.18710/B7BWXO/*"
```

We should fix the paginator to return an empty `next` when you hit the last accessible page of the results. Additionally, we could maybe consider adding a message somewhere that there are more results available and the user should retry with cursor-based pagination.

PR #833, opened 2022-07-01
Pagination fix

digitaldogsbody:
## Purpose
Fixes invalid `next` links being returned when using page-based pagination. This was caused by three different patterns:

1. Removing `next` when the result set is empty - meaning that the last page of results will still have a `next`, and it will only be removed on page(max + 1)
2. Removing next when the size of the results on that page is < `page_size` - this works only when the number of results is not equally divisible by the page size (e.g it works ok for 10 results in pages of 4, because page 3 will have only 2 results and so pass the `if results.size < page[:size]` check, but it will fail for 10 results in pages of 5, because page 2 will have 5 results and so fail that check)
3. When number of results is >10k, pagination calculation would always indicate another page was available, even though it was then clamped by `page[:number].to_i > 0 ? [page[:number].to_i, max_number].min : 1`

closes: #828 

## Approach
Fixed by adding an extra test to all controllers that return paginated results, `page[:number] == total_pages`, meaning `next` is not added when the final page of results are hit, regardless of page size or number of results.

Two new tests have been also added to the datacite_dois_spec file to test that `next` links are correctly removed. These complement the existing test above which tests the functionality in (2) above

## Types of changes

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -56,7 +56,7 @@ class ActivitiesController < ApplicationController
         options[:links] = {
           self: request.original_url,
           next:
-            if results.size < page[:size] || page[:size] == 0
+            if results.size < page[:size] || page[:size] == 0 || page[:number] == total_pages
               nil
             else
               request.base_url + "/activities?" +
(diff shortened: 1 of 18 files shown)


jrhoads (approved):
This all looks great for the rest API.
Approving.

AND I think we should take a look at the `pageInfo` for the graphql apis.  Specifically the functions that determine `hasNextPage` and whether it should return and `endCursor`. 

I will open an issue.

jrhoads:
GraphQL pagination issue opened
https://github.com/datacite/lupo/issues/834

richardhallett (approved):

Issue #829, opened 2022-06-10
`fields[]` querystring parameter is lost when using pagination


digitaldogsbody:
The `next` links provided by a REST API query strip out any filter supplied in the `fields` querystring parameter.

E.g [https://api.datacite.org/dois?page[size]=2&page[number]=1&fields[dois]=id&query=doi:10.18710/B7BWXO/*](https://api.datacite.org/dois?page[size]=2&page[number]=1&fields[dois]=id&query=doi:10.18710/B7BWXO/*)

returns a `next` link of `https://api.datacite.org/dois?page[number]=2&page[size]=2&query=doi:10.18710/B7BWXO/*` (url-decoded), and the `fields[dois]=id` part of the request is lost.

PR #835, opened 2022-07-04
Add fields to pagination

digitaldogsbody:
## Purpose
When requesting a limited fieldset (e.g to save data transfer when the user just wants IDs or some other field for each result in a large query), the `fields` parameter can be added. This was not being included in the pagination links and so lost beyond the first page of results.

closes: #829 

## Approach
Adds a new function to the Fieldable concern to return the data in a format that can be added to a querystring hash and includes it for all controllers that can use `fields` parameter. Added RSpec tests to the `/dois` request suite, and a new suite for the concern (previously without any tests at all)

## Learning
For the tests, the fact that one hash came from serialized data and one constructed meant the keys did not match as they were strings not symbols. Resolved by using [HashWithIndifferentAccess](https://apidock.com/rails/ActiveSupport/HashWithIndifferentAccess)

## Types of changes

### DIFF
--- a/app/controllers/clients_controller.rb
+++ b/app/controllers/clients_controller.rb
@@ -111,6 +111,7 @@ class ClientsController < ApplicationController
                 "page[number]" => page[:number] + 1,
                 "page[size]" => page[:size],
                 sort: params[:sort],
+                fields: fields_hash_from_params(params)
               }.compact.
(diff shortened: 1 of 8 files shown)


jrhoads (approved):
👍

PR #855, opened 2022-08-29
Prefix bug - moving prefix assignment to lupo create repository funct…

svogt0511:
## Purpose

closes: datacite/bracco#639

Upgrade the client model to assign to a prefix after successful client creation.  This functionality is being removed from Fabrica. 

The other part of this work is datacite/bracco#639.  It is just removal of prefix assignment from Fabrica on repository creation.  

This code should not be merged to master until the Fabrica changes are ready.

## Approach

In Lupo, the general approach is:

- Add validation to the client model that tries to ensure that there are prefixes available before client creation.

- Add 'after_create' hook that on successful client creation, procures a prefix and assigns it to the client.

- Available prefixes: Check the client's associated provider for available provider-owned prefixes (state = 'without-repository').  If none are available, an available prefix (state = 'unassigned') is taken from the general prefix pool.

- Returns an error if there are no prefixes available.

This is combined with removal of the functionality from bracco.

See [here](https://github.com/datacite/bracco/pull/649) for 'TESTING and DEPLOYMENT'.

#### Modifications to the Lupo Test Suite

Lupo tests make various assumptions about the environment settings and data for test runs.  With auto-assignment of prefixes, many of these assumptions must be changed.  The following changes to environmental set up are provided as necessary:
  
- Many tests that created clients were originally failing due to lack of available prefixes.  A pool of available prefixes is now provided for each 'it'/example/test.  (See spec/rails_helper.rb).

- Some tests originally assigned prefixes manually to clients.  They were changed as necessary to take advantage of the auto prefix assignment that is now provided by the client model instead of manual prefix assignment.

- Some tests originally assigned a specific prefix to a client at creation time and depended on that specific prefix being used for other operations.  With auto prefix assignment in the client model, prefix assignment is dynamic, so  those tests were modified as necessary to assign additional prefixes if a specific prefix is required for testing.

- Lazy 'let' evaluation had to be changed in some cases to 'let!' to assure that some values were available before they were used.

- Lupo issue #859: ElasticSearch indices were not always being reset as needed to allow consistent indices to be provided to the test environment, causing test failures. Tests have been modified as necessary to set test metadata 'elasticsearch: true' to assure that  consistent ES indices are provided for each test.

        it "returns error message", elasticsearch: true do ...

#### Managing the Test Prefix Pool

The prefix pool for tests is established in [rails_helper.rb, here](https://github.com/datacite/lupo/blob/0dc0d847ceacf2d52c866e25e024a5d0b8e2f771/spec/rails_helper.rb#L79).  By default a set of 20 free prefixes is provided for every test.  There are some test suites that require a prefix pool size of at least 20 free prefixes.  

The number of prefixes in the pool can be set globally with the environment variable 'PREFIX_POOL_SIZE'.  The size of the prefix pool can also be set on a per-test basis using metadata:

    it "returns error message", prefix_pool_size: 30 do ...

A test can be run with no prefix pool by specifying ':skip_prefix_pool' in the metadata:

    it "returns error message", :skip_prefix_pool do ...    

#### Additional Tests for the Added Functionality:

There are several new tests which have been added to test prefix assignment in the client model.  They are in [clients_spec.rb, line 206](https://github.com/datacite/lupo/blob/0dc0d847ceacf2d52c866e25e024a5d0b8e2f771/spec/requests/clients_spec.rb#L206).

#### Test Flakiness:

Some of the test flakiness that we have observed in the past is due to the RSpec pseudo-randomization of the test suite. It seemed that tests would pass in local development, but fail intermittently in the GitHub test workflow.  It was difficult, if not impossible to reliably reproduce errors seen in the test environment.  It turns out that the pseudo-random behavior can be reproduced in the local dev environment by running the test suite with the same seed value, and so test failures can be reliably reproduced and debugged locally.  For ex:

    bundle exec rspec --seed 20049 spec/models

Additional test flakiness was caused by the ElasticSearch issue, above.

## Types of changes


app/controllers/repositories_controller.rb
```diff
@@ -198,12 +198,32 @@ def show
...
 
   def create
-    @client = Client.new(safe_params)
+    prefix = nil
 
+    @client = Client.new(safe_params)
     authorize! :create, @client
 
-    if @client.save
+    provider_prefix = @client.provider.provider_prefixes.select { |_provider_prefix| _provider_prefix.state == "without-repository" }.first
```
kjgarza, line 206: @svogt0511  Maybe it is safer to do the prefix lookups ahead of the `Client.new(safe_params)` . that way, you know if there are prefixes available before you create any resource. 

And in the case that no prefixes are available, you can send an error right away without creating anything.

app/controllers/repositories_controller.rb
```diff
@@ -198,12 +198,32 @@ def show
...
+        provider_prefix = ProviderPrefix.new({ "provider_id": @client.provider.symbol, "prefix_id": prefix.uid })
+        provider_prefix.save
+      end
+
+      client_prefix = ClientPrefix.new(
+        client_id: @client.symbol,
+        provider_prefix_id: provider_prefix.uid,
+        prefix_id: provider_prefix.prefix.uid,
+      )
+      client_prefix.save
```
kjgarza, line 223: I do think that more test for the controller would be needed. https://github.com/datacite/lupo/blob/master/spec/requests/clients_spec.rb

richardhallett (changes requested):
This code should really be in the model, the model should determine if it's possible to save a client i.e. with a new prefix and return errors as appropriate.
I would think the best course of action would be to actually have a new validate check that does the prefix lookup to determine if one can be assigned. i.e. https://github.com/datacite/lupo/blob/master/app/models/client.rb#L75

Then you'll want something like an after_create callback (i guess because it's new repos only right?) that actually assigns the prefix.

svogt0511:
@richardhallett and @kjgarza I moved the change to the client model if you want to take a look.  Definitely the right way to do it.  I am looking at test failures now.

svogt0511:
@richardhallett and @kjgarza - I have made some changes to the code in client.rb to only check for free prefixes in the validation and then to actually do the assignment in the 'after create' hook.  I also updated the test suite to fix testing problems.  Please take another look at the pull request changes.  Thanks.

svogt0511:
@kjgarza and @richardhallett - I have addressed review comments, added 3 tests [here](https://github.com/datacite/lupo/blob/0dc0d847ceacf2d52c866e25e024a5d0b8e2f771/spec/requests/clients_spec.rb#L206).  The test suite passes and it has been rebased to the head of master.  

Please take another look and if you approve it we can schedule a release.  Thanks.

spec/requests/repositories_spec.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+require "pp"
```
jrhoads, line 4: Can probably get rid of this.

jrhoads (approved):
Other than the one comment about an extra `require pp`
All looks good.

spec/requests/media_spec.rb
```diff
@@ -167,7 +170,7 @@
         }
       end
 
-      it "updates the record" do
+      xit "updates the record", :skip_prefix_pool do
```
kjgarza, line 173: @svogt0511 is this going to stay commented?

spec/requests/repositories_spec.rb
svogt0511: oops.  yes.  I will get rid of that.

spec/models/prefix_spec.rb
```diff
@@ -11,7 +11,7 @@
   end
 
   describe "methods" do
-    it "prefixes all" do
+    it "prefixes all", :skip_prefix_pool do
```
richardhallett, line 14: Shouldn't this be unaffected by any changes, as this is testing Prefixes not their relations?

spec/models/prefix_spec.rb
```diff
@@ -22,7 +22,7 @@
       # expect(meta).not_to be_empty
     end
 
-    it "prefixes with where year" do
+    it "prefixes with where year", :skip_prefix_pool do
```
richardhallett, line 25: Same with this, as there is no need to be caring about a prefix pool

spec/requests/datacite_dois_spec.rb
```diff
@@ -4113,9 +4113,10 @@
     # end
   end
 
-  describe "GET /dois/get-dois", vcr: true do
-    let(:prefix) { create(:prefix, uid: "10.5438") }
-    let!(:client_prefix) { create(:client_prefix, prefix: prefix, client: client) }
+  describe "GET /dois/get-dois", :skip_prefix_pool, vcr: true do
```
richardhallett, line 4116: Instead of needing to skip prefix pool, cant the test be changed to just use one of the ones created?

richardhallett (commented):
I'm not sure I understand the need for the skip prefix pool flag. In the examples used (I commented on a couple but not all) most can be changed to just use the auto assigned prefixes, or in the case of prefix model, shouldnt make any difference.

I understand that the tests have before used a specific prefix, but I'm not sure that's needed in most cases, it's more testing other functionality which just needs a prefix to exist (which can come from the one assigned to the repos).

In general I'd really rather avoid having a specific flag that does something different, even if it is just in the tests.

Otherwise all looks okay.

spec/requests/repositories_spec.rb
svogt0511: Removed "require pp".

spec/requests/media_spec.rb
svogt0511: Done.

spec/models/prefix_spec.rb
svogt0511: I will change it.  I was getting things based on the assumption there were no other prefixes available.  For example, this expected create_list to end up with 10 prefixes when it ended up with 30.  Then it expected the prefix to be 10.50821 and got 10.5081.  The :skip_prefix_pool was there to allow the tests to run as usual, and I believe they were just as valid.

However, I also think you are correct and we should just test with the prefix pool.  It should be unaffected.

Removing :skip_prefix_pool.

spec/requests/datacite_dois_spec.rb
svogt0511: I am getting rid of any other ':skip_prefix_pool's that I can find.

richardhallett (approved):
With the removal of skip_prefix and change of other tests, I think this makes sense to me now.
Thanks!

PR #863, opened 2022-09-22
Parallel tests

jrhoads:
## Purpose
Run tests in parallel on CI

## Approach
Uses the parallel_tests gem to break rspec tests into groups.
Then uses a matrix in the github action to run each group in parallel

## Types of changes


Gemfile.lock
```diff
@@ -624,7 +626,7 @@ GEM
     zeitwerk (2.6.0)
 
 PLATFORMS
-  ruby
+  x86_64-linux
```
digitaldogsbody, line 629: Will this (and the `google-protobuf` change on L264) cause any problems for people developing on other systems? (are any of us developing outside of a Linux environment anyway? maybe this doesn't matter)

digitaldogsbody (approved):
Looks good - I am very happy to have faster testing!

One question - do the non-GHA workflow changes (i.e the use of `#{ENV['TEST_ENV_NUMBER']}` in the model files and the change in config/database.yml) affect how we need to run tests locally or do we just continue with `bundle exec rspec`?

Gemfile.lock
svogt0511: Good point about google-protobuf.  Interesting approach to speeding up the testing.  I am all for faster testing too.  These took close to an hour when I ran them serially.

jrhoads:
We can still run rspec single threaded as we have.
`bundle exec rails db:setup RAILS_ENV=test` and then `bundle exec rspec`

If we want to take advantage of parallel tests locally we can run
`bundle exec rails parallel:setup RAILS_ENV=test` and then `bundle exec rails parallel:spec`

So, we one could change their test behavior...or not...as suites their style.

jrhoads:
In practice, I'll probably run it single threaded as I am developing and just re-running one test or a single file.
But when I go to test the full suite, I'll switch over to parallel testing.

svogt0511:
@jrhoads - I am assuming that multi-threading has no effect on the ES indices or the DB as the tests make assumptions about the existing test data.

jrhoads:
If you run it locally, it will set up multiple separate dbs and multiple
separate ES indices.  One for each thread.

svogt0511 (approved):

PR #864, opened 2022-09-23
Refactor workflows to reuse common components

jrhoads:
## Purpose
Many workflows need to run tests, we should run tests the same way in each place.
We can call one workflow from within another.

## Approach
Refactor parallel tests into a workflow that can be called from within other workflows.
Same with rubocop workflow
Create a pull_request workflow that uses both of these.

## Types of changes

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,94 +4,11 @@ on:
     branches:
       - "master"
 jobs:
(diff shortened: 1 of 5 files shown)


digitaldogsbody (approved):
Great work - this will make it a lot easier to keep everything in sync if/when we need to make changes to the workflows in future

Issue #834, opened 2022-07-01
Paginaiton for GraphQL APIs


jrhoads:
The Graphql apis tend to return a true value for `hasNextPage` even when there are no more pages.

Ex:
The Query
https://commons.datacite.org/doi.org?query=worms+AND+velvet&published=2017
has 41 results total.  The first 25 are returned on the first page and the `pageInfo` has a `hasNextPage` value of true (expected and correct) and a `endCursor`

Using the cursor to navigate to the next page
https://commons.datacite.org/doi.org?query=worms+AND+velvet&published=2017&cursor=MTUxMjU3NTk0OTAwMCwxMC4xNTQ2OC9kbC5tYnhsamw

We get the remaining 16 results but the `pageInfo` returns a `hasNextPage` value of `true` and an `endCursor`.
This is wrong.  And when the user navigates to the next page they get a "No works found" message.

Likely area of code
https://github.com/datacite/lupo/blob/3d73de9e98b37c294c061318a059910e2fbede9a/app/graphql/connections/elasticsearch_model_response_connection.rb#L139

PR #867, opened 2022-09-30
Improve logic for has_next_page for all ElasticSearch based Graphql endpoints

jrhoads:
## Purpose
GraphQL endpoints currently have logic that almost always returns true for hasNextPage.
This is an improvement on that logic but is not perfect.

closes: #834

## Approach
If the size of the array of nodes returned from the search is the same as the size specified by the `first` parameter then it is likely that there is another page.

There is a possibility that there won't be a next page if the `total_count.modulo(@first_value) ==0`. and it is the last page of results.  This does not account for  that but it is better than always returning true.

## Learning
This is a difficult problem
https://discuss.elastic.co/t/how-to-check-if-es-has-more-documents-when-paginating-with-search-after-api/151419

## Types of changes

### DIFF
--- a/app/graphql/connections/elasticsearch_model_response_connection.rb
+++ b/app/graphql/connections/elasticsearch_model_response_connection.rb
@@ -137,7 +137,7 @@ class ElasticsearchModelResponseConnection
 
   # @return [Boolean] True if there are more items after this page
   def has_next_page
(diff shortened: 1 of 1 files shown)


digitaldogsbody (approved):
I agree with Richard. Seems like a knotty problem and this improves existing behaviour.

PR #877, opened 2022-11-17
Doi enrichment - Add subjects to repositories/clients

jrhoads:
## Purpose
Adds subjects to the Client/Repository model.
Connects this with the repository REST API.

closes: https://github.com/datacite/datacite/issues/1549

## Learning
https://medium.com/@DrawandCode/how-and-why-json-schema-in-rails-always-saves-the-day-60ecc68ab303

## Types of changes


app/models/client.rb
```diff
@@ -188,6 +195,14 @@ class Client < ApplicationRecord
       indexes :deleted_at, type: :date
       indexes :analytics_dashboard_url, type: :text
       indexes :cumulative_years, type: :integer, index: "false"
+      indexes :subjects, type: :object, properties: {
+        subjectScheme: { type: :keyword },
+        subject: { type: :keyword },
+        schemeUri: { type: :keyword },
+        valueUri: { type: :keyword },
+        lang: { type: :keyword },
```
kjgarza, line 203: this approach using JSON schema seems nice and interesting.

However, do you really need space in the index for storing language and classificationCode. As thought this mapping is just to map repository subjects to resource subjects. Using the whole set of properties looks more like a storing metadata to describe a subject.

jrhoads: Since the REST API largely relies on the values in the ES index, it seems like it would be a good idea to include these.
Also, this is directly copied from the practice associated with indexing the subjects for the DOIs.

I don't think it hurts to do so.

Gemfile.lock
```diff
@@ -383,10 +388,6 @@ GEM
       connection_pool (~> 2.2)
     netrc (0.11.0)
     nio4r (2.5.8)
-    nokogiri (1.13.9-arm64-darwin)
-      racc (~> 1.4)
-    nokogiri (1.13.9-x86_64-darwin)
-      racc (~> 1.4)
```
digitaldogsbody, line 389: Probably nothing in this - just trying to work out why it's happening. Is it just because the previous updater of Gemfile.lock was on Mac and you're not? Or if we need to ship all three as options does that mean that Bundle can install the correct version for the architecture it's being run on (as opposed to trying to install all three and failing)?

If the latter, maybe we can leave these in?

app/models/client.rb
digitaldogsbody: +1 for JSON schema - I'm happy to see this as a PoC after our chat about it before. I agree that using the values here for driving the serializer is better, as otherwise there might have to be database lookups, which could cause an N+1 queries situation.

digitaldogsbody (approved):

Gemfile.lock
jrhoads: I have a feeling that this was here (or absent) because the previous updater was on mac and I am not.
For production and Docker the only one needed is the Linux one.  The rest could stay or go without much fuss.

My understanding is that Bundler is smart enough to ignore the versions that it doesn't need.

PR #886, opened 2022-12-08
Refactor safe params to make a more skinny controller

kjgarza:
## Purpose
Refactor controller to a more `skinny controller` with less non-response-related logic. Making the code in safe_params testable.
Reduces code complexity for [DOI controller and safe_params method](https://codeclimate.com/github/datacite/lupo/issues?category%5B%5D=complexity&status%5B%5D=&status%5B%5D=open&status%5B%5D=confirmed&language_name%5B%5D=Ruby).

## Approach

- Replace Function with Command.
- Moving non-response-related logic and non-persistence-related logic, to a domain object.
- Using `app/lib` to take advantage of autoloading.
## Learning

- Concerns can override many methods names. `Crosscitable` is a dangerous concern as it brings loads of methods from bolognese all of which have very common names that can cause overriding method errors.

## Types of changes


app/controllers/datacite_dois_controller.rb
```diff
@@ -778,441 +777,26 @@ def safe_params
...
-          contentUrl: [],
-          identifiers: [],
-          relatedIdentifiers: [],
-          relatedItems: [],
-          fundingReferences: [],
-          geoLocations: [],
-        },
-      }
+      ParamsSanitizer.sanitaize_nameIdentifiers(params[:creators])
+      ParamsSanitizer.sanitaize_nameIdentifiers(params[:contributors])
```
digitaldogsbody, line 781: Small typo in the function name here

app/controllers/datacite_dois_controller.rb
```diff
@@ -778,441 +777,26 @@ def safe_params
...
-        :downloadCount,
-        :partCount,
-        :partOfCount,
-        :referenceCount,
-        :versionCount,
-        :versionOfCount,
-        :viewCount,
-      )
+    def sanitized_params
+      ParamsSanitizer.new(safe_params.to_h).cleanse
```
digitaldogsbody, line 799: If `safe_params` should always return a hash, the `.to_h` might be redundant

app/lib/params_sanitizer.rb
```diff
@@ -0,0 +1,461 @@
...
+  def add_schema_version(hash)
+    @params[:schemaVersion] =
+      if METADATA_FORMATS.include?(hash["from"])
+        LAST_SCHEMA_VERSION
+      else
+        @params[:schemaVersion]
+      end
+  end
+
+  def self.sanitaize_nameIdentifiers(array)
```
digitaldogsbody, line 451: Small typo in function name

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
+        expect(subject.add_xml).to eq(nil)
+      end
+    end
+
+    context "when xml is string" do
+      let(:params) { { "xml" => "<xml></xml>" } }
+      subject { ParamsSanitizer.new(params) }
+      it "should return hash" do
```
digitaldogsbody, line 45: I think the test name here might be wrong, since `add_xml` will return a string or nil

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        expect(subject.add_xml).to eq(nil)
+      end
+    end
+  end
+
```
digitaldogsbody, line 50: Can you add some more tests here for the various other things that this function is expected to do?

From reading the code, it decodes the base64 encoded `xml` and returns it, stripping the BOM and any whitespace, so it would be good to have tests for:

- Does a correctly encoded b64 string return a correct string of the contents?
- Does a badly encoded b64 string (e.g incorrect padding) return nil?
- Does the BOM get stripped correctly if present?
- Does any whitespace get stripped correctly?

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      let(:params) { { "version_info" => "" } }
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
+        subject.add_metadata_version(params)
+        expect(subject.get_params[:version_info]).to eq(nil)
+      end
+    end
+
+    context "when params is not hash" do
+      let(:params) { { "version_info" => 123 } }
```
digitaldogsbody, line 72: I think that params _is_ a hash here so this context should maybe be renamed

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
```
digitaldogsbody, line 74: This description does not match the assertion, I think it should be "should return correct value" or similar

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      let(:params) { { "prefix" => "", "doi" => "" } }
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
+        subject.add_random_id
+        expect(subject.get_params[:doi]).to eq(nil)
+      end
+    end
+
+    context "when params is not hash" do
+      let(:params) { { "prefix" => 123, "doi" => 123 } }
```
digitaldogsbody, line 103: params is hash here - and I think this test is actually testing for when the prefix is not valid so perhaps renaming would make things clearer?

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
+        subject.add_random_id
+        expect(subject.get_params[:doi]).to eq(nil)
+      end
+    end
+
+    context "when params is hash" do
```
digitaldogsbody, line 111: Likewise here, this test is actually for when a prefix is valid, so renaming might be useful

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      subject { ParamsSanitizer.new({ prefix: "10.1233", doi: "" }) }
+
+      it "adds correctly random doi" do
+        subject.add_random_id()
+        expect(subject.get_params[:doi]).to start_with("10.1233")
```
digitaldogsbody, line 116: I think because the test environment generates predictable DOIs (per comment and code here: https://github.com/datacite/lupo/pull/886/files#diff-1f6e1cee6b8f670353834f39818e744200a5edc30f509ed1655ac918b7295a0cR430) we could increase this assertion to the whole DOI rather than just the prefix

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    context "when params is empty" do
+      let(:params) { { "schemaVersion" => "" } }
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
+        subject.add_schema_version(params)
+        expect(subject.get_params[:schemaVersion]).to eq(nil)
+      end
+    end
+
+    context "when params is not hash" do
```
digitaldogsbody, line 142: Another case where params is hash, and the test is for whether or not valid data is correctly returned - so renaming could be useful

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+      let(:params) { { "schemaVersion" => 123 } }
+      subject { ParamsSanitizer.new(params) }
+      it "should return nil" do
```
digitaldogsbody, line 145: The title here does not match the assertion below

spec/lib/params_sanitizer_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        subject.add_schema_version(params)
+        expect(subject.get_params[:schemaVersion]).equal?(123)
+      end
+    end
+
+    context "when params is hash" do
```
digitaldogsbody, line 151: I think this is actually testing whether the `from` parameter works properly, so renaming to make that clear might be helpful

app/lib/params_sanitizer.rb
```diff
@@ -0,0 +1,461 @@
...
+      @params[:schemaVersion],
+    ].compact
+  end
+
+  def get_params
+    @params
+  end
+
+  def add_xml
+    # extract attributes from xml field and merge with attributes provided directly
```
digitaldogsbody, line 364: Now that things have been refactored, this comment is no longer accurate as the merging happens outside of this function, so it could maybe be reworded to reflect that or removed

digitaldogsbody (changes requested):
This looks great - I think refactoring the code like this will definitely be helpful for reducing complexity going forwards :)

My main question is do you think it might be worth expanding the test suite a bit? Since the refactoring is happening partially in order to make testing easier, it seems like an opportune moment to expand the functionality of the tests and make sure they cover all aspects of the code.

I have added a couple of suggestions about this in comments on the test suite file, but there are also some functions of the new `ParamsSanitizer` class that are not covered at all (e.g `add_xml_attributes` and `sanitize_nameIdentifiers`) and I think it could be good to add coverage now if possible.

Last request - could you write up the reasoning behind this refactor and methodology into an ADR? I think it would be really useful to have as a reference for future refactoring efforts - thanks!

digitaldogsbody:
Also, just to say, I spent some time going over this with Cody today. It's a little out of the areas of Ruby/Rails knowledge for both of us, so it might also be helpful for @richardhallett and @jrhoads to take a look at the code before merge :)

app/controllers/datacite_dois_controller.rb
kjgarza: good catch many thanks :thumbsup:

app/controllers/datacite_dois_controller.rb
kjgarza: Actually, this is necessary because of strong params. Otherwise, you will be cloning the params object. this is practically one of the reasons safe params is so big.

spec/lib/params_sanitizer_spec.rb
kjgarza: I think that would be extra

app/controllers/datacite_dois_controller.rb
digitaldogsbody: Thanks - the strong params part was the bit of knowledge I was missing here

kjgarza:
@codycooperross @digitaldogsbody 

> there are also some functions of the new ParamsSanitizer class that are not covered at all (e.g add_xml_attributes and sanitize_nameIdentifiers) and I think it could be good to add coverage now if possible.

`sanitize_nameIdentifiers` is a class method, there is already a test in the controller for that. 
I have added a few more tests for the other methods. Remember that most of these are currently tested by proxy in integration tests. 

hopefully, with this refactoring bugs will be easier to catch or new unit tests can be created for each case.

kjgarza:
@digitaldogsbody I could add an ADR, but I am not sure if this refactoring merits that. it seems to me more of a case of not using basic `the single-responsibility principle when coding **safe_params**. 

But if we needed it one it might look like [this](https://gist.github.com/kjgarza/864a1a4e4b47f56720eb225c72a21798)

svogt0511 (commented):
This PR looks really good to me.  We can always add more tests.  I ran the entire test suite locally and it passes.

svogt0511 (approved):

PR #887, opened 2022-12-13
Adds title sort by first title functionality to the /dois endpoint

codycooperross:
## Purpose

Adds title sort by first title functionality to the /dois endpoint to fulfill datacite/datacite#1518

## Approach

The spec called for title sorting by the DOI's first title. Because Elasticsearch has no sense of order in object/non-nested fields, we were unable to select the first value of an array in a sort using the expected sort method:

``` 
        { "titles.title.keyword": { order: "asc" } }
```

We attempted to use a Painless script to sort by first title, but the response time and unpredictable system load made this option appear untenable: 

```
        { "_script": {
            "type": "string",
            "script": {
                "lang": "painless",
                "inline": "if (params._source.titles.length > 0) { return params._source.titles[0].title } else { return '' }"
            },
            "order": "asc"
          }
        } 
```

Instead, I created a new attribute in the Doi model `primary_title` that selects the first title from the Doi's `titles`. This is wrapped in an array for interoperability with other applications that expect titles to be in an array. The `primary_title` attribute is added to the indexed JSON and is sortable using the implemented sort. 

- [x] For the `primary_title` attribute to be mapped added in ES and the sort to be implemented, DataCite DOIs will need to be re-indexed.  

## Learning

We tested the first, non-cached load of the /dois endpoint locally using multiple ES title sort methods. The [Elasticsearch guide on "Scripts, caching, and search speed"](https://www.elastic.co/guide/en/elasticsearch/reference/current/scripts-and-search-speed.html) indicated that the only ways to improve search speed for a larger index with script sorts is to create an ingest pipeline with script processor and reindex. Rather than manage the additional suggested attribute with a script processor in ES, it seems better to consolidate the attribute logic in the Doi mode. 

## Types of changes

### DIFF
--- a/app/controllers/datacite_dois_controller.rb
+++ b/app/controllers/datacite_dois_controller.rb
@@ -44,9 +44,25 @@ class DataciteDoisController < ApplicationController
       when "-citation-count"
         { citation_count: { order: "desc" } }
       when "title"
-        { "titles.title.keyword": { order: "asc" } }
+        { "_script": {
+            "type": "string",
+            "script": {
+                "lang": "painless",
+                "inline": "if (params._source.titles.length > 0) { return params._source.titles[0].title } else { return '' }"
+            },
+            "order": "asc"
+          }
+        }
       when "-title"
-        { "titles.title.keyword": { order: "desc" } }
+        { "_script": {
+            "type": "string",
+            "script": {
+                "lang": "painless",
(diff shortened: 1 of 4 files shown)


jrhoads (approved):
Looks good.
Yes, this will require a DOI re-index.

PR #893, opened 2022-12-27
Adjusts validation of Doi language attribute to allow non-ISO 639-1 values

codycooperross:
## Purpose

Removes validation of the Doi model's language attribute to allow non-ISO 639-1 free-text values. Allows for free-text entry in the Language property per datacite/datacite#1590

## Approach

I removed the language attribute validations in the Doi model and moved the logic to a computed attribute. ISO 639-1 codes and ISO 639-1 "name" values (like "English") are still normalized to a ISO 639-1 code. Free-text values that are neither ISO 639-1 codes or ISO 639-1 "name" values are saved as submitted. 

- [ ] The XSD uses a `xs:language` type for the language property. Registered/findable metadata can produce the following errors if the language attribute contains commas, dashes, and/or other special characters. Permitting any text may cause unexpected validation errors because the attribute text had been previously normalized to a ISO 639-1 code or `nil`.

```{
    "source": "language",
    "uid": "10.14454/z5bs-0327",
    "title": "DOI 10.14454/z5bs-0327: 'Borgesian, Spanish' is not a valid value of the atomic type 'xs:language'. at line 15, column 0"
}
```

## Types of changes

### DIFF
--- a/spec/models/doi_spec.rb
+++ b/spec/models/doi_spec.rb
@@ -298,18 +298,17 @@ describe Doi, type: :model, vcr: true, elasticsearch: true do
       expect(doi.language).to eq("fr")
     end
 
-    it "non-iso 639-1" do
+    it "error" do
       doi.language = "hhh"
       expect(doi.save).to be true
       expect(doi.errors.details).to be_empty
-      expect(doi.language).to eq("hhh")
+      expect(doi.language).to be_nil
     end
 
     it "nil" do
       doi.language = nil
       expect(doi.save).to be true
       expect(doi.errors.details).to be_empty
-      expect(doi.language).to be_nil
     end
   end
 
--- a/spec/models/doi_spec.rb
+++ b/spec/models/doi_spec.rb
@@ -298,8 +298,29 @@ describe Doi, type: :model, vcr: true, elasticsearch: true do
       expect(doi.language).to eq("fr")
     end
 
-    it "error" do
+    it "human longer than 8 characters" do
+      doi.language = "Indonesian"
+      expect(doi.save).to be true
+      expect(doi.errors.details).to be_empty
+      expect(doi.language).to eq("id")
+    end
+
+    it "non-iso 639-1" do
       doi.language = "hhh"
+      expect(doi.save).to be true
+      expect(doi.errors.details).to be_empty
+      expect(doi.language).to eq("hhh")
+    end
+
+    it "non-iso 639-1 with country code" do
+      doi.language = "prq-PE"
+      expect(doi.save).to be true
+      expect(doi.errors.details).to be_empty
+      expect(doi.language).to eq("prq-PE")
+    end
+
+    it "fails xs:language regex" do
+      doi.language = "Borgesian"
       expect(doi.save).to be true
       expect(doi.errors.details).to be_empty
       expect(doi.language).to be_nil
--- a/spec/models/doi_spec.rb
+++ b/spec/models/doi_spec.rb
@@ -298,17 +298,18 @@ describe Doi, type: :model, vcr: true, elasticsearch: true do
       expect(doi.language).to eq("fr")
     end
 
-    it "error" do
+    it "non-iso 639-1" do
       doi.language = "hhh"
       expect(doi.save).to be true
       expect(doi.errors.details).to be_empty
-      expect(doi.language).to be_nil
+      expect(doi.language).to eq("hhh")
     end
 
     it "nil" do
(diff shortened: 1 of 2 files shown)


codycooperross:
## Purpose


Adjusts validation of the Doi model's language attribute to allow values that conform to the DataCite Metadata Schema's language property xs:language type restrictions. Allows for user entry in the Language property per datacite/datacite#1590

## Approach


ISO 639-1 codes and ISO 639-1 "name" values (like "English") are still normalized to a ISO 639-1 code. Values that are neither ISO 639-1 codes or ISO 639-1 "name" values are saved as submitted if they conform to the xs:language type regex. Values that cannot be normalized and that do not match the xs:language type regex continue to be saved as nil, reflecting the existing behavior.

## Types of changes

richardhallett:
Why do we want to allow anything that is not ISO 639, isnt xs:language saying it should always be an ISO code, and not whatever user decides? Perhaps I'm misunderstanding the intent.

kjgarza:
> Why do we want to allow anything that is not ISO 639, isnt xs:language saying it should always be an ISO code, and not whatever user decides? Perhaps I'm misunderstanding the intent.

Yes, and the intention is Not to change that. 
Fair warning I have not reviewed this PR. 
Currently, the API and schema are expecting an ISO code, but the API is discarding anything that doesn't fit 639-1 (mind you is not erroring; it is just discarding it from ES, see https://github.com/datacite/bracco/pull/690#pullrequestreview-1229624846 ). So the intent is that when the user sends any other lang-code, it's stored in ES, provided that it doesn't fail the language directive in the schema.

So the only behavior change intended is that when there is no error, then the value is stored. But i have not review this PR so i don't know if this change accomplishes that intent

codycooperross:
> Currently, the API and schema are expecting an ISO code, but the API is discarding anything that doesn't fit 639-1 (mind you is not erroring; it is just discarding it from ES, see [datacite/bracco#690 (review)](https://github.com/datacite/bracco/pull/690#pullrequestreview-1229624846) ). So the intent is that when the user sends any other lang-code, it's stored in ES, provided that it doesn't fail the language directive in the schema.

Just a note that, previously, non-ISO 639-1 values were discarded from the _database_ and consequently ES. One of the original tests for language attempts to save a non-ISO 639-1 value in language, which is expected to be stored in the database as `nil`. In the PR, this behavior is maintained for languages that fail the language directive in the schema. In other words, invalid language submissions will continue to be saved as `nil` instead of causing save errors. 

> Why do we want to allow anything that is not ISO 639, isnt xs:language saying it should always be an ISO code, and not whatever user decides? Perhaps I'm misunderstanding the intent.

The xs:language type actually allows for a wider arrange of codes, as defined here: https://www.w3.org/TR/xmlschema11-2/#language This PR maintains normalization for ISO 639-1 codes and language "names" (ex. "English") but allows entry of other valid codes as well.

spec/models/doi_spec.rb
```diff
@@ -298,10 +298,31 @@
...
+      doi.language = "prq-PE"
+      expect(doi.save).to be true
+      expect(doi.errors.details).to be_empty
+      expect(doi.language).to eq("prq-PE")
+    end
+
+    it "fails xs:language regex" do
+      doi.language = "Borgesian"
+      expect(doi.save).to be true
+      expect(doi.errors.details).to be_empty
```
digitaldogsbody, line 325: If the language saved is `nil` (i.e the input was invalid), is there a reason we are not returning an error message?

digitaldogsbody (approved):

spec/models/doi_spec.rb
codycooperross: I'm not sure. This was how it was implemented before—see the original `it "error" do` test, which is expected to return `nil` instead of an error—so I imagine there was some intention. My guess is that we didn't want to reject all records with non-ISO 639-1 codes and names in the language property since this could cause headaches for users.

digitaldogsbody: I am worried that silently dropping data is not a good idea, even if it's what we have already been doing.

codycooperross: That makes sense. That said, I am concerned with changing the behavior, since this may cause issues with records that could have been submitted successfully in the past. The XSD and the model validation will block any invalid values even if they make it in.

codycooperross: The current behavior has been logged as an issue here: https://github.com/datacite/lupo/issues/904

Issue #849, opened 2022-08-16
"dates.dateInformation" is not searchable via ElasticSearch


codycooperross:
### Describe the bug

The `dates.dateInformation` field is not searchable in ElasticSearch. This affects searches in the REST API and Commons.

#### Expected Behaviour

The `dates.dateInformation` field should be searchable.

#### Current Behaviour

Queries like `dates.dateInformation:*` return 0 results despite the existence of of DOIs like 10.4224/40000277, which contain `dateInformation` fields: https://api.datacite.org/dois?query=doi:10.4224/40000277

#### Steps to Reproduce

1. Search https://api.datacite.org/dois?query=dates.dateInformation:*

#### Hypothesis

Maybe due to the lack of `dateInformation` ES mapping here: 

https://github.com/datacite/lupo/blob/f0d6078fd378c8b609e59d9bc51bf4d560492013/app/models/doi.rb#L306

Issue #907, opened 2023-01-19
`fundingReferences.schemeUri` not searchable in REST API / ES


digitaldogsbody:
Same cause as issue #849, the `schemeUri` property is missing from the relevant ES index definition: https://github.com/datacite/lupo/blob/master/app/models/doi.rb#L300-L306

digitaldogsbody:
Fixed in https://github.com/datacite/lupo/pull/909

Issue #906, opened 2023-01-19
`relatedItems.contributor.name` not searchable in REST API / ES


digitaldogsbody:
The index template has the field incorrectly titled as `contributorName`.

This is what it is in the XML, but the Bolognese `datacite_reader` uses a shared `get_authors` function when converting it to JSON: https://github.com/datacite/bolognese/blob/master/lib/bolognese/readers/datacite_reader.rb#L241

This eventually calls a function that gets either the `creatorName` or `contributorName` depending on which one exists on the element and standardises it to `name`: https://github.com/datacite/bolognese/blob/b0a7df3c9dd6a45eaf56fd0e06d304e4db9b837d/lib/bolognese/author_utils.rb#L22-L23

All other ES index definitions use `name`, so the fix should be to also apply this to the related items contributor section of the index definition on https://github.com/datacite/lupo/blob/master/app/models/doi.rb#L286

digitaldogsbody:
Fixed in https://github.com/datacite/lupo/pull/909

PR #909, opened 2023-01-19
Fix ES index definition to match the JSON data structure

digitaldogsbody:
## Purpose

Fixes various inconsistencies between the ES index mapping and the way the JSON data that populates it is generated by Bolognese

closes: #849, #906, #907 

## Approach

Update the mappings in the DOI model

Post-merge & deploy - this will require a re-templating of the `dois` index in ES to generate a new mapping, and not just a reindex.

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -283,7 +283,7 @@ class Doi < ApplicationRecord
         edition: { type: :keyword },
         contributors: { type: :object, properties: {
           contributorType: { type: :text },
-          contributorName: { type: :text },
+          name: { type: :text },
           nameType: { type: :text },
           givenName: { type: :text },
           familyName: { type: :text },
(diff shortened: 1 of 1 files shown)


digitaldogsbody:
Rubocop about to fail me for bad spacing in brackets...

richardhallett (approved):
Think it makes sense, please check indexes as expected on staging.
Also did this not break any existing tests?

digitaldogsbody:
> Also did this not break any existing tests?

No because the data is already in this format in ES (as it is what Bolognese generates) and so the fields (and data within) are already returned when you pull the ES records, and the issue is just that the ES mapping for the index didn't match, so those fields were not accessible directly.

codycooperross (approved):

PR #912, opened 2023-01-24
Add analytics_tracking_id to repositories model

richardhallett:
## Purpose
As part of the new Analytics service we need a way of specifying a tracking id used to identify the analytics data for repositories collected.

This adds a new field alongside the existing analytics_dashboard_url to the model, it then appears in the various serialisers.
A new permission has been added so only providers/clients can work with this.

Note that this also does remove analytics_dashboard_url from the existing clients_serializer as this is not necessary to appear here.

Also while analytics_dashboard_url is being unused at present, I've decided to leave it in the model, while this is slightly messy it potentially does have future usage, and/or we may wish to do a further cleanup operation of the model.
This also makes this a relatively easy change to deploy as we can just index the new field.

## Approach
Add new model
Add permissions can can
Add tests related to permissions
Add request test to check data exists

To reviewers, I added most of you for visibility.

Nada

## Learning
N/A

## Types of changes


jrhoads (approved):
Looks Good

PR #914, opened 2023-01-25
Explicitly set the is_write_index key on every switch_index option

digitaldogsbody:
## Purpose

Currently when running the `datacite_doi:switch_index` task and _only_ when moving from `dois_v2` to `dois_v1`, the `is_write_index` flag is not set due to faulty logic in the Indexable concern. This prevents ES from indexing any new DOIs, causing significant errors with REST API functionality.

## Approach

Explicitly sets `is_write_index` in both branches of the `switch_index` rake task logic.

## Types of changes

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -1099,6 +1099,7 @@ module Indexable
               add: {
                 index: index_name,
                 alias: alias_name,
+                is_write_index: is_write_index,
               },
             },
           ],


jrhoads (approved):

digitaldogsbody:
Steps to recreate the issue locally:

1. Manually assign the `dois` alias to `other-dois` index on your local ES (using Elasticvue)
2. From the indices page look at the properties of `dois_v1` and `dois_v2` (from the cog icon->show info) - the active one should have `"is_write_index": true` in the `aliases` json block at the top
3. If currently active is v2, skip to step 5
4. SSH into the local Lupo container and run `bundle exec rake datacite_doi:switch_index` to switch the active index to v2
5. Refresh Elasticvue and confirm that `dois_v2` now has `"is_write_index": true` in its properties
6. SSH into the local Lupo container and run `bundle exec rake datacite_doi:switch_index` to switch the active index (back) to v1
7. Refresh Elasticvue and see that `dois_v1` does not get `"is_write_index": true` in its properties
8. Confirm that `dois_v2` is also missing this (as it should be right now, because it's the inactive index) 
9. Confirm that the active `dois-other` index which shares the `dois` alias is also missing this (as it should because it will never be the write alias for `dois`) **THIS IS THE POINT WHERE NONE OF THE INDEXES BEHIND THE `dois` ALIAS ARE CONFIGURED AS THE WRITE INDEX** - From now on, any attempt to index a DOI to ES will cause an error
10. Run the switch_index task again to switch back to v2, and confirm that the `"is_write_index": true` is now present and indexing DOIs is now possible again

PR #916, opened 2023-02-02
Doi enrichment - Index DOIs with client/repository subject area classifications

jrhoads:
## Purpose
Index doi with subject information from the parent repository.

closes: https://github.com/datacite/datacite/issues/1550
closes: https://github.com/datacite/datacite/issues/1747

## Approach
At index time, include information on the subjects from the parent repository.
Create new index fields for the parent repository(client) subject information
Create new index fields for combined (doi and repository) subject information

In the GraphQL endpoint for works
Display field and facet information for fieldOfScience based on the parent repository subject information
Display field and facet information for fieldOfScience based on the combined subject information

## Learning
Json objects can be mapped to the index in two different ways. Either as a flat `object` or as a `nested` field.
In order to have complicated filters/facets *at search time* based on other fields within an the Json object (Such as the SubjectScheme) then a nested field is required.

An alternative is to create new mappings that pre-filter the subject based on the scheme *at index time*.
This is what we have chosen to do here for fieldOfScience facets.  This has the added benefit that if a facet is selected for a filter later, the filter parameters are pretty clear and need to only reference one field. 

## Types of changes

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -198,7 +198,7 @@ class Client < ApplicationRecord
       indexes :analytics_dashboard_url, type: :text
       indexes :analytics_tracking_id, type: :text
       indexes :cumulative_years, type: :integer, index: "false"
-      indexes :subjects, type: :object, properties: {
+      indexes :subjects,  type: :nested, include_in_parent: true,  properties: {
         subjectScheme: { type: :keyword },
         subject: { type: :keyword },
         schemeUri: { type: :keyword },
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -923,7 +923,6 @@ class Client < ApplicationRecord
           "Subjects are only allowed for disciplinary repositories.  This repository_type is: #{repository_type}",
         )
       end
-
     end
 
     def check_id
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -396,6 +396,7 @@ class Client < ApplicationRecord
         end,
       "analytics_dashboard_url" => analytics_dashboard_url,
       "analytics_tracking_id" => analytics_tracking_id,
+      "subjects" => Array.wrap(subjects),
     }
   end
 
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -510,7 +510,7 @@ class Client < ApplicationRecord
   end
 
   def subjects=(value)
-    write_attribute(:subjects, Array.wrap(value))
+    write_attribute(:subjects, Array.wrap(value).uniq)
   end
 
   def opendoar=(value)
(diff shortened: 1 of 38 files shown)


spec/factories/client.rb
```diff
@@ -18,6 +18,29 @@
     role_name { "ROLE_DATACENTRE" }
     password_input { "12345" }
     is_active { true }
+    subjects do
+      [
+        {
+          classificationCode: "1001",
+          schemeUri: "http://example.com/schemeUri",
+          subject: "Example Subject",
+          subjectScheme: "Example Subject Scheme (ESS)",
```
digitaldogsbody, line 27: Given that this is explicitly _not_ FOS, is it worth having slightly different values for the code and subject text so that they can be asserted against individually in tests without needing to also assert against the subjectScheme field?

spec/graphql/types/work_type_spec.rb
```diff
@@ -1201,4 +1201,93 @@
...
+      response = @facet_response
+      expect(
+        response.dig("data", "works", "fieldsOfScienceCombined")
+      ).to match_array([
+        {
+          "id" => "example_subject",
+          "title" => "Example Subject",
+          "count" => WORK_COUNT
+        }
+      ])
```
digitaldogsbody, line 1290: I might be getting the wrong end of the stick - but it looks like the data being tested here only comes from one half of the combined source (the repository). Would it be worth adding another subject record to the work, so that there is one coming from either part of the combination, and testing against that to ensure that the data is getting combined correctly?

app/models/doi.rb
```diff
@@ -408,6 +410,14 @@ class Doi < ApplicationRecord
         updated: { type: :date },
         deleted_at: { type: :date },
         cumulative_years: { type: :integer, index: "false" },
+        subjects: { type: :nested, include_in_parent: true,  properties: {
```
digitaldogsbody, line 413: Now that the approach is to index FOS separately, could this be switched to being object type rather than nested?

app/models/client.rb
```diff
@@ -198,7 +198,7 @@ class Client < ApplicationRecord
       indexes :analytics_dashboard_url, type: :text
       indexes :analytics_tracking_id, type: :text
       indexes :cumulative_years, type: :integer, index: "false"
-      indexes :subjects, type: :object, properties: {
+      indexes :subjects,  type: :nested, include_in_parent: true,  properties: {
```
digitaldogsbody, line 201: Likewise here, could this be switched back to being an object type?

digitaldogsbody (approved):
Great work - I have added a couple of questions about the indexing and the test suite, but none of them are blockers to merge :)

app/models/doi.rb
jrhoads: Yes

app/models/client.rb
jrhoads: Will do

Issue #920, opened 2023-02-09
Repository creation via Fabrica is very slow


svogt0511:
### Describe the bug

Creating a repository in Fabrica using the form takes 7 - 10 seconds for the 1 api call to create a repository.  

#### Expected Behaviour

The repository is created and returns a 200 response in reasonable amount of time and successfully allocates a prefix to that repository.

#### Steps to Reproduce

Create a repo using the form in Fabrica.

#### Hypothesis

This is related to the complexity in of implementing the relationships between repositories and prefixes in Rails Active Model, in addition to the size of the DB tables and the number of calls that must be made to choose a free prefix.  Elastic search is much faster and seems to work well for this purpose.

Changed this code to use elastic search instead of directly querying the DB to get the status of prefixes.

Issue #918, opened 2023-02-03
Repository created via Fabrica may return a 409 error containing "The resource already exists."


svogt0511:
### Describe the bug

When creating a repository using Fabrica, the API may return a 409 error reading "The resource already exists."  See the screen shot.

![image](https://user-images.githubusercontent.com/4681304/216722505-fa6d9a33-b363-44eb-9a10-130197503fec.png)

#### Expected Behaviour

The repository is created and returns a 200 response.

#### Steps to Reproduce

The error is returned unexpectedly when creating a repository account in Fabrica. The form does not respond on submission.

I have not seen this happen in Fabrica in staging.  I have seen it using Fabrica in test, but I have also seen repositories created successfully in test and production, so the error seems to happen inconsistently.

#### Hypothesis

The error may be related to obtaining a prefix for the user upon repository creation.

Issue #917, opened 2023-02-03
Repository creation via Fabrica may return a 400 error response containing a NoMethodError 


codycooperross:
### Describe the bug

When creating a repository using Fabrica, the API may return a 400 error reading "undefined method `analytics_tracking_id' for #<Client:0x00007fe7b7cb9b10> Did you mean?  analytics_dashboard_url". The client will be created but the error response will cause Fabrica not to redirect appropriately. 

#### Expected Behaviour

The repository is created and returns a 200 response.

#### Steps to Reproduce

The error is returned unexpectedly when creating a repository account in Fabrica. 

#### Hypothesis

The error may be related to user abilities affecting the RepositorySerializer.

PR #921, opened 2023-02-09
Fix issues relating to automatic prefix allocation while creating repositories.

svogt0511:
## Purpose

Closes: #920, #917, #918.  The last 2 issues are related and it will close those, as well.  They are all due to one issue located in the repositories#create controller, where the API request to create the repo would take a very long time, and eventually Fabrica would time out on that request, but not report success or failure to the user.   Additionally, the API request would eventually be successfully completed, but that would never be reported back to the Fabrica user.

There were errors reported in the browser dev console (accounts for errors as reported in  #917 and also spurious CORS errors seen in the console after the API request timed out).  When the user would try again to create the repository, the form would report the 409 error, "resource already exists" (accounts for #918).  

This is also related to Bracco issue datacite/bracco#721 where form errors aren't reported back to the user.  However, this pull request and the one associated with datacite/bracco#721  can be released separately.

## Approach

Found a Rails 6.1 ActiveRecord feature that produced a query for getting a free prefix that was much faster than the original one.  Benchmarked the old query and the new query in staging and found that the new query is much faster (about 350 times faster than the old one).  That is sufficient for our purposes.

DECIDED AGAINST THE ORIGINAL APPROACH:
Uses ElasticSearch to look for free prefixes.  Refactors the code which will hopefully speed things up and fix bugs.  

## Types of changes

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -922,15 +922,8 @@ class Client < ApplicationRecord
       end
     end
 
-    def get_prefix
-      provider_prefix = (provider.present? && provider.provider_prefixes.present?) ? provider.provider_prefixes.select { |_provider_prefix| (_provider_prefix.state == "without-repository") }.first : nil
-      prefix = Prefix.all.count > 0 ? Prefix.all.select { |_prefix| (_prefix.state == "unassigned") }.first : nil
-
-      provider_prefix || prefix || nil
-    end
-
     def check_prefix
-      if !get_prefix
+      if !prefixes_available
         errors.add(
           :base,
           "No prefixes available.  Unable to create repository.",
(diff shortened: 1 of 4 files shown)


richardhallett (commented):
What's the reason for switching to ES to do lookup?, its such a tiny amount of data the query to the DB shouldnt be that slow.
Did you benchmark the slowness? Is it definitely from the DB?

With this you've now got the scenario where available prefixes has a delay before stuff is indexed, as opposed to going to DB which will ensure that the changes are updated immediately and therefore the available prefixes will be up to date.
This avoids multiple checks and also prevents needing that sleep in the test factory which will slow down testing.

svogt0511:
Richard,


You would think that it is a tiny amount of data that it would be really quick. I looked at the API requests to create a repository from Fabrica in the dev console, networking tab. I assumed that it was the prefix lookup that took so long since it was much quicker before I added that. It could have been something else, and not the DB request. 


I am going to retrace my steps and get back to you with better evidence about what is causing the problem.


Your last point is one thing that had me worried.


Thanks. I will be in touch.


Suzanne

--- original message ---
On March 21, 2023 at 2:43 PM EDT ***@***.*** wrote:



@richardhallett commented on this pull request.

What's the reason for switching to ES to do lookup?, its such a tiny amount of data the query to the DB shouldnt be that slow.

Did you benchmark the slowness? Is it definitely from the DB?

With this you've now got the scenario where available prefixes has a delay before stuff is indexed, as opposed to going to DB which will ensure that the changes are updated immediately and therefore the available prefixes will be up to date.

This avoids multiple checks and also prevents needing that sleep in the test factory which will slow down testing.

—

Reply to this email directly, view it on GitHub, or unsubscribe.

You are receiving this because you were assigned.Message ID: ***@***.***>
--- end of original message ---

svogt0511:
@richardhallett - I logged into a client-api server thru the bastion and, in the rails console, ran a function with that timed the line of code that finds an unassigned prefix.  3.6 seconds.  Definitely a very expensive query (or set of queries) to the DB.  What I am going to do is look for something less expensive as an alternative to using elastic search.  Will let you know when I find it and you can take a look.

digitaldogsbody:
> @richardhallett - I logged into a client-api server thru the bastion and, in the rails console, ran a function with that timed the line of code that finds an unassigned prefix. 3.6 seconds. Definitely a very expensive query (or set of queries) to the DB

Is Rails doing some "magic" and making a series of JOINs like prefix->client->dois and causing much more data than necessary to be retrieved?

richardhallett:
> > @richardhallett - I logged into a client-api server thru the bastion and, in the rails console, ran a function with that timed the line of code that finds an unassigned prefix. 3.6 seconds. Definitely a very expensive query (or set of queries) to the DB
> 
> Is Rails doing some "magic" and making a series of JOINs like prefix->client->dois and causing much more data than necessary to be retrieved?

Hmm maybe but looking at the the models, I think it's actually doing multiple table scan for every single row, because provider_prefix does not contain a "state", it calculates value based on client_prefixes existing. So ye super slow.

Then maybe ES is the solution here, tbh I think the DB could be remodelled to make this just as easy but for this it might just be simpler to go ahead with ES.
Another option or perhaps in combination might just to put the prefix assignment on the queue, so repos get created but prefix assignment happens "later" in practical terms this will most likely be within a minute.

svogt0511:
@richardhallett and @digitaldogsbody - I updated the query.  The new query is much faster - about 350 times faster than the old one, on average.  Good enough for this purpose.  I tested it in staging.  When I was done, I moved the master branch back to staging.  If you want to test there, we can redeploy this branch to staging.  

Please review it again when you can.

digitaldogsbody (approved):
New query looks good

richardhallett (approved):
Ah left outer join, yes makes sense, it is basically same except if I was to guess the previous was doing multiple queries as state has to be calculated.
Id have thought maybe merge would be slow as in memory but I guess resultset is pretty small.
I could probably write a SQL query that'd do it in one, but not sure how to translate that into ActiveRecord ORM speak.
Anyway if tests work out then seems good to me

digitaldogsbody:
Is Rails not smart enough to turn the `.merge` into either an `AND` or a `UNION` clause? That was my assumption for how it would work.

svogt0511:
@richardhallett, @digitaldogsbody - Thanks for the reviews. The amount of result data is small so I am not worried that will slow things down. It complicated things that the state is calculated from more than one table.  Also, it is probably faster if left to MySQL, but in staging, this was much better than the original - on average 380 times faster than the original which turned seconds into milliseconds. Probably slower in production, but still acceptable.

Some of the things we can do in the future: we can supply our own queries through ActiveRecord instead of letting ActiveRecord do the translation.   Additionally, there is a plugin that will spit out the SQL query formulated by ActiveRecord. That may be useful in development.  And perhaps some instrumentation is in order - there are benchmarking add-ons.  My testing consisted of running a script to time the active record query in rails console.  While perfectly sufficient for this purpose, we may need more in the future.  

Also, I tried looking through the 'slowqueries' log in AWS Cloudwatch.  Very difficult to find things.  Also, there are only logs for production.  I noticed the DB error logs coming through to DataDog.  And someone set up a dashboards in DataDog for both active record and for db queries which look like they might be useful.

Thanks again!

PR #933, opened 2023-03-16
Refactor GH Actions

jrhoads:
## Purpose
This PR refactors in the deploy and release action to use a common build action.
It also implements steps to tag the image created in the deploy action with a main tag.

These are in preparation for further work that will have a manual deploy trigger for non-main/master branches to staging.

## Types of changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,58 @@
+name: Build and Tag
+on:
+  workflow_call:
+    secrets:
+      DOCKERHUB_USERNAME:
+        required: true
+      DOCKERHUB_TOKEN:
+        required: true
+      GITHUB_TOKEN:
+        required: true
+    inputs:
+      image_name:
+        description: The name of the image for docker/ghcr registries
+        required: true
+        type: string
+      image_tag:
+        description: Tag for the image for docker/ghcr registries
+        required: true
+        type: string
+
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Cache Docker layers
+        uses: actions/cache@v3
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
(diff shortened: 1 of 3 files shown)


svogt0511 (approved):
Just a couple of questions.  I see the build.yml can be invoked manually with whatever image_tag you specify (really the branch name).  Then deploy.yml always sets the image_tag to MAIN for the build_and_push, which seems fine for now.  I see that release.yml sets the image_tag to whatever ref_name is and I am assuming that it will always be whatever the tag is when released from the main page of the repo.  If that is the case, It seems fine.  

How did you test this?
Are we getting rid of slack notifications?
Ultimately will we be able to pull images for staging that we can now push via build.yml?

Otherwise.  Looks good to me.

jrhoads:
@svogt0511 Thanks for taking a look.

`build.yml` can be invoked from within another action.  In this case it is called from within `deploy.yml` or `release.yml`.
This is what the `on: workflow_call` trigger does.

There's another trigger `on: workflow_dispatch` which I think of as more of a manual trigger.  I have a couple of later PRs that will use this manual trigger to dispatch non-main/master branches.

Yes.  For release tags the `ref_name` is always the tag.  I used this page extensively to figure out what information was available through the github context variables. https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

I've been using `act` for local testing to help make sure the syntax is correct.

The Slack job/notifications in GH actions have been failing for a while now.  They are removed from some of our other projects.  The github bot in slack does most of our notifications for PRs/issues/releases.  If we want to update the Slack job in the future that would be fine, but since it hasn't been working for a while, I just removed it.

This PR behaves the same as what we have right now.  A future PR will introduce the ability to `build` images for staging and then `update_terraform` so that the images get deployed into the staging environment.

svogt0511:
> This PR behaves the same as what we have right now. A future PR will introduce the ability to `build` images for staging and then `update_terraform` so that the images get deployed into the staging environment.

Thanks for the reply confirming my my understanding how the changes work.  I wanted to make sure I understood that correctly.  

Also thanks for confirming that, even with these changes the behavior hasn't changed, yet.

I look forward to the end result (being able to test branches other than main in staging).

Nice to know about 'act'.  I will have to try it out again.  I resorted to testing these things using my private git environment.

Looks good.

digitaldogsbody (approved):
Great work - I think that rationalising the workflows like this gives us a much better process. I am looking forward to having this in action.

PR #962, opened 2023-05-15
Openapi-schema

kjgarza:
serving openapi schema directly from the container to allow for CORS during intern work


sarala:
@kjgarza, I see this has gone through without a review. It would be great if all pull requests are reviewed. I have assigned @digitaldogsbody and @jrhoads to review. Could you also please add some text around what we will be using this for.

sarala:
I just assigned @richardhallett as Joseph and Mike are out this week.

richardhallett:
Yes what is this? I dont understand the reason this is required.

GraphQL spec is automatically generated it is not a fixed schema. Where is the file being generated from?
Our graphql endpoint is api.datacite.org/graphql

Also what is "/.well-known/ai-plugin.json"?

PR #975, opened 2023-06-22
Fair Islands Visualization Feature 1 - Getting all scholarly ouputs linked to an organization

jrhoads:
## Purpose
Ensure that works are indexed to be associated with an organization if it fullfills any of the following scenarios:

1. Direct work( Org is creator): creator.nameIdentifier == Organization.rorID
2. Association through creator affiliation
3. Org is direct sponsor of contributor
4. Association through contributor affiliation
5. Association through DMP

As outlined in https://github.com/datacite/datacite/issues/1824

## Approach
This PR sets up the indexing for these indices.
It **also** implement the filtering itself.

Closes https://github.com/datacite/datacite/issues/1824

## Types of changes


bklaing2 (approved):
Looks good to me

app/models/doi.rb
```diff
@@ -1808,8 +1813,47 @@ def consortium_id
...
+      Array.wrap(c.fetch("nameIdentifiers", nil)).each do |name_identifier|
+        sum << ror_from_url(name_identifier.fetch("nameIdentifier", nil)) if name_identifier.is_a?(Hash) && name_identifier.fetch("nameIdentifierScheme", nil) == "ROR" && name_identifier.fetch("nameIdentifier", nil).present?
+      end
+      sum
+    end
+  end
+
   def affiliation_id
-    (Array.wrap(creators) + Array.wrap(contributors)).reduce([]) do |sum, c|
+    (Array.wrap(creators) + sponsor_contributors).reduce([]) do |sum, c|
```
digitaldogsbody, line 1856: Is the `affiliation_id` property used anywhere else that might be accidentally impacted by this change? e.g when building a non-GQL query based on querystringparams supplied to the /dois endpoint

app/models/doi.rb
```diff
@@ -1819,7 +1863,7 @@ def affiliation_id
   end
 
   def affiliation_id_and_name
-    (Array.wrap(creators) + Array.wrap(contributors)).reduce([]) do |sum, c|
+    (Array.wrap(creators) + sponsor_contributors).reduce([]) do |sum, c|
```
digitaldogsbody, line 1866: Same query as above for this property

app/models/doi.rb
```diff
@@ -1808,8 +1813,47 @@ def consortium_id
     client.provider.consortium_id.downcase if client.present? && client.provider.consortium_id.present?
   end
 
+  def related_dmp_ids
+    related_identifiers.select { |related_identifier|
+      related_identifier["relatedIdentifierType"] == "DOI"
+    }.select { |related_identifier|
+      related_identifier.fetch("resourceTypeGeneral", nil) == "OutputManagmentPlan"
```
digitaldogsbody, line 1820: Think there is a typo here, per the [Schema XSD for ResourceTypes](https://schema.datacite.org/meta/kernel-4.4/include/datacite-resourceType-v4.xsd) it should be `OutputManagementPlan`

digitaldogsbody (changes requested):
Requesting one change for a typo, other than that it looks good.

Would be good to understand more about the usages of these properties (see the couple of questions) but this is not a blocker to merge :)

app/models/doi.rb
jrhoads: This should be resolved

digitaldogsbody (approved):
Looks good to me :)

jrhoads:
Turns out that the filtering was a two line addition and then an additional test.
It will work more completely after the reindex but will not throw errors in the mean time.

app/models/doi.rb
richardhallett: Yes this seems less flexible now we're only explicitly looking at Sponsor contributors, whereas before we could search for any contributor for the affiliation id.

jrhoads: Y'all are probably right.  I'll revert this and create a new field.

jrhoads: In the engineering meeting, we discussed and explored this a bit.  While affiliation_id is present in the REST api, the logs indicate that it is only being used in the graphql api.  In the documentation, the fields behind affiliation_id are not explicitly defined; the implementation could change.

I will go ahead with the change and keep an eye for support issues to see if it creates indirect effects.

kjgarza: I fully support your initial instinct to create a new field, @jrhoads. It might seem convenient in the short term to repurpose the existing affiliation_id field for this new feature, but in the long term, it can have unintended effects on the integrity and usability of the API.

We need to remember that the functionality you're implementing is specific to the FAIR Island use case. Therefore, it shouldn't introduce changes that impact the general usage of the other DataCite services. Users should be able to continue interacting with the APIs in the same way they always have.

Our focus should be on preserving the current user experience as much as possible while extending the functionality where needed for FAIR Island. In this light, creating a new field would provide a much cleaner and safer implementation. It enables users who want to use the FAIR Island definition to do so explicitly, such as in the FAIR Island dashboard, without affecting those who don't need it.

Building on that point, it's worth remembering that each component of our system should ideally have one responsibility, one reason to change. By repurposing the affiliation_id field, we're essentially asking it to do two things: its original function, and a new function for the FAIR Island functionality. This goes against the grain of clean, maintainable code. We should aim to create features that cut across all the necessary layers of our architecture (UI, business logic, data access) but within a narrow scope. In this case, the scope is the FAIR Island functionality. It makes sense to isolate the change within this scope, rather than letting it bleed into other areas of the system.


Let's prioritize minimizing the disruption to the existing user experience over short-term convenience.

PR #997, opened 2023-08-16
Refactor DOI GraphQL Queries

jrhoads:
- Remove print statement
- Refactor Doi.gql_query into a module using the builder pattern
- Refactor the Doi:GraphqlQuery::Builder further into smaller functions
- Add spec for GraphqlQuery::Builder
- Appease rubocop

## Purpose
The Doi model is very large with many concerns.
This PR is to refactor the creation of the query into a class to make it more readable and more testable.

## Types of changes


app/models/doi/graphql_query.rb
```diff
@@ -0,0 +1,355 @@
...
+    DEFAULT_CURSOR = [0, ""]
+    DEFAULT_PAGE_SIZE = 0
+    DEFAULT_FACET_COUNT = 10
+
+    def initialize(query, options)
+      @query = query
+      @options = options
+    end
+
+    def search_query
```
richardhallett, line 16: Maybe explicitly name t his like: build_full_search_query
This possibly then gives the notion of building a query without everything listed.

app/models/doi/graphql_query.rb
```diff
@@ -0,0 +1,355 @@
...
+            query: clean_query,
+            fields: query_fields,
+            default_operator: "AND",
+            phrase_slop: 1
+          }
+        }]
+      end
+    end
+
+    def filters
```
richardhallett, line 117: Part of me wonders if each filter should be a seperate function.
e.g.
```ruby
def add_filter_provider_id
[{ terms: { provider_id: options[:provider_id].split(",") } } ]
```

So then you'd do the options check where you combine all the filters. This could have the benefit that you could build a query via composing the functions, if that makes sense?

richardhallett (approved):
In general I approve, but I added some comments about possible changes that you could think about.
Maybe it might be the case this is a starting point and further work can be done in another PR.

app/models/doi/graphql_query.rb
jrhoads: I have some ideas for that (Specifically around aggregations).  For this refactor I wanted to keep things as close as the original functionality.

I think the next steps would be to add more tests for filters, and then look at refactoring the filters.

digitaldogsbody (approved):
Great work! Let's get this merged :)

At some point we should have a chat about the GQL filters because there's a bit left over from 4.4 regarding the new resourceTypeGeneral elements added and how they interact (or don't) with GQL because of the way the filtering is currently designed.

PR #1023, opened 2023-10-18
Set reconnect: true in config/database.yml to avoid random timeouts du…

svogt0511:
…ring db queries.

## Purpose

Reconnect on DB query timeouts.  This is related to the timeout that happened during a db migration operation in staging using the Departure gem (w/ percona toolset).  The  migration was completed but on updating the schema_migrations table to show that the migration had been done, the query timed out.  This setting would allow the mysql2 client to reconnect and try the op again.

UPDATE - What are the side-effects of allowing reconnects like this.  Is it safe?  Would increasing the timeout period be a good alternative solution?

## Types of changes

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -11,6 +11,7 @@ defaults: &defaults
   database: <%= ENV['MYSQL_DATABASE'] %>
   host: <%= ENV['MYSQL_HOST'] %>
   port: <%= ENV['MYSQL_PORT'] %>
+  reconnect: true
 
   <<: *mysql
 


wendelfabianchinsamy:
Some further reading suggest that this feature is deprecated in the most recent versions of MySQL https://dev.mysql.com/doc/c-api/8.0/en/c-api-auto-reconnect.html#:~:text=If%20auto%2Dreconnect%20is%20enabled%2C%20mysql_ping()%20performs%20a%20reconnect,used%20to%20suppress%20this%20behavior.

Furthermore, the reconnect option is not transaction safe and rolls back any active transactions and resets the autocommit mode (removes the implicit start transaction...commit statement around each statement that is not part of a transaction).

More info and side effects can be found on the link I've provided.

@svogt0511 thank you for getting me to pause and think about this for a bit.

jrhoads:
From @wendelfabianchinsamy comment, it seams like this PR is a no-go.
Should we go ahead an close and approach the problem another way? @svogt0511

richardhallett:
Ye seems like not the right way to go about this. I had thought this was a rails reconnect flag, not mysql specific.
I don't think the timeout is a huge concern for the rollout of the new column, if we check it was successfully manually.

svogt0511:
Closing this without merging due to the [discussion in todays Development group meeting](https://docs.google.com/document/d/18GEhuWryhs-irA1rl19fLr3oHI84tCnfacAwoYTbaSE/edit#bookmark=id.fji14d3c9no3) and the above comments.  It was the general consensus that using the 'reconnect' option to mysql was not safe, and with the departure gem, probably unnecessary.

PR #1026, opened 2023-10-20
GraphQL person search errors if there is not a query value

wendelfabianchinsamy:
## Purpose
closes: https://github.com/datacite/datacite/issues/1949

## Approach
Add a null safe operator the when stripping out the URL scheme from the the query value. The query can, it seems, sometimes be null. The previous implementation was simply passing the null value into the GraphQL query so we will continue with that approach.

## Types of changes


bklaing2 (approved):
Looks good, thanks!

PR #1027, opened 2023-10-23
Changes to run lupo locally which will be using other docker services…

ashwinisukale:
… like elastic search,mysql and memcach

## Purpose

Setting up the Lupo application locally, without relying on Docker services, enhances the development process by increasing productivity. In the past, we often had to delete the image, recreate it, and restart Docker services to observe the latest code changes.

With the local application setup, we now have the convenience of swiftly testing both Bolognese and Lupo modifications. This approach saves a significant amount of time that would otherwise be spent regenerating the Docker image to reflect the latest code.

closes: https://github.com/datacite/lupo/issues/1029

## Approach

We will just remove the web service from the docker compose file and we will use docker for other services like Elasticsearch,  Memcached and MySql.
We will call these services from the public port of the docker on local system. 
You need to do some modification in local environment file, for the port and host. We will use public port of the docker services here.

Note: Sometime `localhost` does not work so try with `127.0.0.1`

Use following values in `.env` file on local system

```
ES_HOST=localhost:9201
ELASTIC_PASSWORD=changeme
MEMCACHE_SERVERS=localhost:11211
MYSQL_HOST=localhost
MYSQL_PORT=3309
```
## How to Run
This will start all other services like Elasticsearch, Memcached and MySql on your local system
```
docker-compose -f docker-compose.dev.yml up
```

## Types of changes


richardhallett (changes requested):
Could we potentially rename the file to be docker-compose.depends.yml or something similar?

For new developers .dev might suggest this is the only way that docker instances can be spun up, but in reality the main docker-compose file is a way for development to happen.

digitaldogsbody:
I agree with Richard's comments, making it clear that this is an alternate method is a good idea

ashwinisukale:
Thanks @richardhallett how about `docker-compose.local.dev.yml`

richardhallett:
Sure that's fine.

ashwinisukale:
Thanks @richardhallett I have addressed the review comment. Could you please review it again.

richardhallett (approved):

digitaldogsbody:
> In the past, we often had to delete the image, recreate it, and restart Docker services to observe the latest code changes.

It occurred to me the other day, that with `RAILS_ENV` set to development, the deployed local container _should_ watch for changes to the code and hot-reload, as the `/app` directory is mounted to the container by docker-compose, rather than just copied in.

If you are using a rails console inside the container, you can use the `reload!` command to bring in any changes to the code since the console opened.

richardhallett:
Thanks Mike, I should have mentioned this as well, this is how I work with lupo, everything is hot reloaded.
I have encountered the need to rebuild docker containers but usually only as a result of changing the container config or in some instances an odd caching instance, the **reload!** command is a great tip too.

I do think it's fine if you want to work without being inside a container though, depending on your machine docker can be an overhead.

PR #1052, opened 2023-12-06
Allow events to trigger on specific conditions

ashwinisukale:
## Purpose

Events not created when DOI first created in registered or draft state

closes: https://github.com/datacite/datacite/issues/1710

## Approach

Events should be created at following conditions,

- When we create a DOI directly with `findable` state
- When state is `findable` and there is an update in any of these fields `related_identifiers`, `creators`, `funding_references`
- When we change DOI state from `any other state` to `findable`

## Types of changes

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -21,11 +21,10 @@ module Indexable
       end
 
       if instance_of?(DataciteDoi) || instance_of?(OtherDoi) || instance_of?(Doi)
-        if aasm_state == "findable" && !Rails.env.test?
-          # If following fields are modified or the state has changed (any other state to findable), send import message
-          if saved_change_to_attribute?("related_identifiers") || saved_change_to_attribute?("creators") || saved_change_to_attribute?("funding_references") || aasm_state_changed?
-            send_import_message(to_jsonapi)
-          end
+        if aasm_state == "findable"
+          changed_attributes = saved_changes
+          relevant_changes = changed_attributes.keys & %w[related_identifiers creators funding_references aasm_state]
+          send_import_message(to_jsonapi) if relevant_changes.any?
         end
       elsif instance_of?(Event)
         OtherDoiJob.perform_later(dois_to_import)
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -20,17 +20,12 @@ module Indexable
         end
       end
 
-      if (
-         instance_of?(DataciteDoi) || instance_of?(OtherDoi) ||
-           instance_of?(Doi)
-       ) &&
-          (
-            saved_change_to_attribute?("related_identifiers") ||
-              saved_change_to_attribute?("creators") ||
-              saved_change_to_attribute?("funding_references")
-          )
-        if aasm_state == "findable" && !Rails.env.test?
-          send_import_message(to_jsonapi)
+      if (instance_of?(DataciteDoi) || instance_of?(OtherDoi) ||instance_of?(Doi))
+        if aasm_state == "findable"
+          # If following fields are modified or the state has changed (any other state to findable), send import message
+          if saved_change_to_attribute?("related_identifiers") || saved_change_to_attribute?("creators") || saved_change_to_attribute?("funding_references") || aasm_state_changed?
+            send_import_message(to_jsonapi)
+          end
         end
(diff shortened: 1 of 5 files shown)


spec/requests/old_events_spec.rb
```diff
@@ -915,7 +915,10 @@
   end
 
   context "show" do
-    let(:doi) { create(:doi, client: client, aasm_state: "findable") }
+    let(:doi) do
+      allow_any_instance_of(DataciteDoi).to receive(:send_import_message)
```
jrhoads, line 919: Does this intercept the message in tests so we can remove the guard in the Indexable concern?

spec/models/doi_spec.rb
```diff
@@ -31,6 +31,85 @@
...
+      it "sends import message if creators are modified" do
+        travel_to(Time.zone.local(2023, 12, 14, 10, 7, 40)) do
+          expect(doi).to receive(:send_import_message).with(doi.to_jsonapi)
+
+          doi.update(creators: [{ "nameType" => "Personal", "name" => "New Creator" }])
+        end
+      end
+
+      it "sends import message if funding_references are modified" do
+        travel_to(Time.zone.local(2023, 12, 14, 10, 7, 40)) do
```
jrhoads, line 63: Not a blocker:
I'd be interested in hearing more about this `travel_to` method.

spec/requests/old_events_spec.rb
ashwinisukale: I am not sure if I get your question, but previously we never used to execute `send_import_message` method in test env, but I have removed that condition as I wanted to write a code which checks when we should execute that method.

Hence we need to mock this method call.

https://github.com/datacite/lupo/blob/master/app/models/concerns/indexable.rb#L32

Let me know if I answered your query :)

spec/models/doi_spec.rb
ashwinisukale: Sure, `mock record` value and `actual record` value were giving me same data in expect call, but `updated_at` was showing mili second difference in time, as they created at different time, to avoid that I used time travel.

My test case was failing intermittently, because of time issue,  

```
  -    "updated"=>2023-12-14 09:55:48.000000000 +0000},
  +    "updated"=>2023-12-14 09:55:49.000000000 +0000}
```

With the `travel_to` method, you can "time travel" within your tests or specific parts of your code. It's like telling your program, "Hey, for the next little while, pretend it's a different time."

app/models/concerns/indexable.rb
```diff
@@ -20,17 +20,11 @@ module Indexable
...
-              saved_change_to_attribute?("creators") ||
-              saved_change_to_attribute?("funding_references")
-          )
-        if aasm_state == "findable" && !Rails.env.test?
-          send_import_message(to_jsonapi)
+      if instance_of?(DataciteDoi) || instance_of?(OtherDoi) || instance_of?(Doi)
+        if aasm_state == "findable"
+          changed_attributes = saved_changes
+          relevant_changes = changed_attributes.keys & %w[related_identifiers creators funding_references aasm_state]
+          send_import_message(to_jsonapi) if relevant_changes.any?
```
digitaldogsbody, line 27: I really like this refactor away from an ever-increasing list of boolean clauses, very elegant!

digitaldogsbody (approved):
Great work Ashwini :)

wendelfabianchinsamy (approved):

app/models/concerns/indexable.rb
ashwinisukale: Thank you @digitaldogsbody :)

Also `saved_change_to_attribute?` this method only execute when we have `update` call, so we never captured the `create` DOI call before because of this method checks.

We missed the events for the DOI which are created directly with `findable` state.

ashwinisukale: `saved_changes` will work for both `create` and `update` action.

spec/requests/old_events_spec.rb
jrhoads: Cool

spec/models/doi_spec.rb
jrhoads: This is a great tool to add to our testing toolbelt.  Thanks

jrhoads (approved):

Issue #1073, opened 2023-12-18
Update lupo events update action find event logic


wendelfabianchinsamy:
We used to find the events using the UUID. However, the UUID will not be part of the response body with the new /relationships CrosssRef endpoint.

We will use the { subj_id, obj_id, source_id, relation_type_id } index to find events.

wendelfabianchinsamy:
Waiting on CrossReft to complete work on the new endpoint.

PR #1074, opened 2023-12-18
search for events using sub_id+obj_id+source_id+relation_type_id

wendelfabianchinsamy:
## Purpose

closes: https://github.com/datacite/lupo/issues/1073

## Types of changes

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -46,7 +46,13 @@ class EventsController < ApplicationController
 
   def update
     @event = Event.where(uuid: params[:id]).first
-    exists = @event.present?
+
+    @event = Event.find_by(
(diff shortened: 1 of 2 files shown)


app/controllers/events_controller.rb
```diff
@@ -46,7 +46,13 @@ def create
 
   def update
     @event = Event.where(uuid: params[:id]).first
```
jrhoads, line 48: Do both of these searches get called?

Should the second one get an `if @event.blank?` much like line 58

wendelfabianchinsamy: Heyo @jrhoads. Both of these should certainly not get called. The lookup via UUID will get removed. Sorry I should have made this PR a draft. Wanted to see why the builds are failing 😢

PR #1081, opened 2024-01-08
Added log to verify that method is getting called at right condition

ashwinisukale:
## Purpose

Added logs to investigate the issue.

closes: https://github.com/datacite/datacite/issues/1710

## Types of changes

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -24,7 +24,10 @@ module Indexable
         if aasm_state == "findable"
           changed_attributes = saved_changes
           relevant_changes = changed_attributes.keys & %w[related_identifiers creators funding_references aasm_state]
-          send_import_message(to_jsonapi) if relevant_changes.any?
+          if relevant_changes.any?
+            send_import_message(to_jsonapi)
+            Rails.logger.info "[Event Data Import Message] #{aasm_state} #{to_jsonapi.inspect} send to Event Data service."
+          end
         end
       elsif instance_of?(Event)
         OtherDoiJob.perform_later(dois_to_import)


kaysiz (approved):
@ashwinisukale Looks good. Is there not a way to simulate this locally without logging in prod?

app/models/concerns/indexable.rb
```diff
@@ -24,7 +24,10 @@ module Indexable
         if aasm_state == "findable"
           changed_attributes = saved_changes
           relevant_changes = changed_attributes.keys & %w[related_identifiers creators funding_references aasm_state]
-          send_import_message(to_jsonapi) if relevant_changes.any?
+          if relevant_changes.any?
+            send_import_message(to_jsonapi)
+            Rails.logger.info "[Event Data Import Message] #{aasm_state} #{to_jsonapi.inspect} send to Event Data service."
```
richardhallett, line 29: This will potentially create a lot of log entries in production for every update that is triggering a import.
You could also get this on the other side during the actual import, i.e. levriero.
I think this is fine if you really need this information to debug in production, but if you can check how many logs are being added that'd be good, and if it's a lot (discuss on slack) then we might want to remove it at some point. Log collection does have an associated cost as well as being extra data to parse.

ashwinisukale: Yes @richardhallett I will not be pushing this to production. I just want to test this on staging. I will be removing this log once testing is done. Sorry I did not mentioned this in the description

richardhallett (approved):

ashwinisukale:
@kaysiz locally its not possible to test this integration testing at local where lupo is putting a message on a queue and levriero is reading that message off the queue

Issue #1016, opened 2023-10-02
Copy of Metadata 4.5 - Database/ElasticSearch/Lupo 


codycooperross:
RESOURCES

- [Metadata Schema Release 4.5 Product Spec](https://docs.google.com/document/d/1DC3SJw0W2bdJQeZAbmpbBbNVgJw5Zgq4HRn4ciIQmEI/edit?usp=sharing)
- [Metadata 4.5 - Technical Spec](https://docs.google.com/document/d/1-NW1Ik_2mFFFcy_keduiZe6DjmuyccSLe-4PD4EqXiM/edit?usp=sharing)

This development task includes making DB/ElasticSearch/Lupo mods to allow initial processing of the new publisher_v4 field while maintaining the old field, migrating all DOIs to the new field, then  cleaning up code and unused resources.

MIGRATION TASKS:

- [ ] Database migration script to add the new field. - publisher_v45, using the prescribed structure.
- [ ] Elastic Search mapping - for the publisher_v45 field.
- [ ] Elastic Search re-index.
- [ ] Lupo create/update endpoints including:
    - [ ] Lupo logic to write to both fields on create/update.
    - [ ] Lupo logic to handle publisher=true/false and get the correct data on get request.
    - [ ] Lupo logic to handle serializer.
- [ ] Lupo get endpoint - same for now.

- [ ] Background script to migrate old publisher field to new field publisher_v45.
- [ ] ElasticSearch migration - add new template for the new field, run the ES migration.

POST MIGRATION/CLEAN UP TASKS:
- [ ] Lupo create/update endpoints including:
  - [ ] Remove old code from create/update endpoints
  - [ ] Add code to populate the publisher_v45 field, being sure this works for both JSON input as well as file input (xml blob).
  - [ ] Use the validation gem mentioned by Joseph to validate the JSON in the publisher_v45 field. (activeRecord_json_validator)
  -  [ ] Add the new vocabulary terms in the other fields (ResouceTypeGeneral and RelatedItems)
- [ ] Lupo get endpoint:
  - [ ] Remove the code that searches the publisher field.
  - [ ] Add  code that searches the publisher_v45 field.
  - [ ] Add code to search for the additional controlled vocabulary terms in the other fields.
  - [ ] Get endpoint: Handle the publisher=true/false query.
- [ ] Database:
  - [ ] Database Rails migration to delete publisher field when we are ready. 

TESTING
- [ ] Appropriate tests for each endpoint - create/read/update/get
[...]

PR #1111, opened 2024-02-01
Schema-4.5 cleanup. (DOI model)

svogt0511:
## Purpose

Related to: Issue #1016, [PR#1112](https://github.com/datacite/lupo/pull/1112)

## Approach

STEP 1 - This PR which obsoletes the publisher column from the database by making it an active record virtual attribute.  To preserve backward compatibility, 'publisher' is still accepted by the API as either a string or a hash, but is saved to the DB or read from the DB as publisher_obj.  Also, the API publisher='true' flag still works as documented to return either a string or hash in the response.

STEP 2 - A second PR ([PR#1112](https://github.com/datacite/lupo/pull/1112)) which has the rake task to remove the publisher column from the DB, and contains the new schema.rb file.

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -5,7 +5,7 @@ require "benchmark"
 
 class Doi < ApplicationRecord
   PUBLISHER_JSON_SCHEMA = "#{Rails.root}/app/models/schemas/doi/publisher.json"
-  audited only: %i[doi url creators contributors titles publisher publisher_obj publication_year types descriptions container sizes formats version_info language dates identifiers related_identifiers related_items funding_references geo_locations rights_list subjects schema_version content_url landing_page aasm_state source reason]
+  audited only: %i[doi url creators contributors titles publisher_obj publication_year types descriptions container sizes formats version_info language dates identifiers related_identifiers related_items funding_references geo_locations rights_list subjects schema_version content_url landing_page aasm_state source reason]
 
   # disable STI
   self.inheritance_column = :_type_disabled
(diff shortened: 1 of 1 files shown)


app/models/doi.rb
```diff
@@ -5,7 +5,7 @@
 
 class Doi < ApplicationRecord
   PUBLISHER_JSON_SCHEMA = "#{Rails.root}/app/models/schemas/doi/publisher.json"
-  audited only: %i[doi url creators contributors titles publisher publisher_obj publication_year types descriptions container sizes formats version_info language dates identifiers related_identifiers related_items funding_references geo_locations rights_list subjects schema_version content_url landing_page aasm_state source reason]
```
digitaldogsbody, line 8: Will removing this remove historical data about changes to the field from the `activities` table/index/endpoint?

digitaldogsbody (approved):
Looks good - one question about the impact of the audit change, but it's non-blocking to merge

app/models/doi.rb
svogt0511: Great question.  Deleting the field name from that line just tells the audit module (activities) to not record changes for that field.  It won't erase historical data for publisher or leave those changes out of a call to /activities. 

Changes are kept in the audited_changes column of the audits table.  For reporting changes, activities just grabs audited_changes, but doesn't care what is in it. For recording changes, the publisher column will be gone.

ashwinisukale (approved):
Looks Good

PR #1112, opened 2024-02-02
Schema-4.5 cleanup. (DB migration)

svogt0511:
## Purpose

Related to: Issue #1016, [PR 1111](https://github.com/datacite/lupo/pull/1111)

## Approach

This is the second of 2 tasks needed to remove the original publisher column from the dataset DB table.  The first, [PR#1111](https://github.com/datacite/lupo/pull/1111), changed publisher to a virtual doi model attribute for purposes of calculating publisher_obj, removing the need to save the publisher field to the dataset table in the DB.   This PR contains the rake task to delete the publisher column from the dataset table in the DB.  This can be invoked as:

```
PERCONA_ARGS='--no-drop-old-table' bundle exec rake db:migrate VERSION=20240229195058
```

We are using the departure gem so work will be done on a copy of the table and the old table will be swapped out for the new one when ready.  The optional `--no-drop-old-table` argument allows us to keep the old table around until we are ready to delete it.  

## Types of changes


codycooperross (approved):
Test the migration locally, and all seems well.

PR #1118, opened 2024-02-09
Add/Remove Indexes - Allocator Table

kaysiz:
## Purpose

We have missing and unused indexes on allocator table.

closes: #1117 

## Types of changes

### DIFF
--- a/db/migrate/20240209113122_remove_globus_uuid_index.rb
+++ b/db/migrate/20240209113122_remove_globus_uuid_index.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class RemoveGlobusUuidIndex < ActiveRecord::Migration[6.1]
   def change
     remove_index :allocator, column: :globus_uuid
--- a/db/migrate/20240209113122_remove_globus_uuid_index.rb
+++ b/db/migrate/20240209113122_remove_globus_uuid_index.rb
@@ -1,0 +1,5 @@
+class RemoveGlobusUuidIndex < ActiveRecord::Migration[6.1]
+  def change
+    remove_index :allocator, column: :globus_uuid
+  end
+end
--- a/db/migrate/20240209120111_add_missing_indexes_to_allocator.rb
+++ b/db/migrate/20240209120111_add_missing_indexes_to_allocator.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMissingIndexesToAllocator < ActiveRecord::Migration[6.1]
+  def change
+    add_index :allocator, :deleted_at, name: 'index_allocator_deleted_at'
+    add_index :allocator, :role_name, name: 'index_allocator_role_name'
+  end
+end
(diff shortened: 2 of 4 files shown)


jrhoads (approved):

digitaldogsbody (approved):

db/migrate/20240209120111_add_missing_indexes_to_allocator.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddMissingIndexesToAllocator < ActiveRecord::Migration[6.1]
+  def change
+    add_index :allocator, :deleted_at, name: "index_allocator_deleted_at"
+    add_index :allocator, :role_name, name: "index_allocator_role_name"
+    add_index :allocator, :ror_id, name: "index_allocator_ror_id"
```
wendelfabianchinsamy, line 7: @kaysiz have we confirmed that the queries missing these indices are in fact using these indices in the query execution plan?

digitaldogsbody: Good question Wendel!

I think if Rails is not smart enough to have done this and we have to do it manually, it can be a separate piece of work/PR and doesn't need to be a blocker to merging this.

wendelfabianchinsamy:
@kaysiz something for us to keep in mind, adding additional indices will have two side-effects:

1. We will increase the page size as we have to store additional indices (increased storage - probably less of an issue but I thought I'd add this as a heads up)
2. We could have a detrimental effect on writes since each write will also have to update the indices.

Just thought I'd add these as a comment just so we could keep an eye on any write performance degradation or if we see a uptick in RDS storage usage.

kaysiz:
> @kaysiz something for us to keep in mind, adding additional indices will have two side-effects:
> 
> 1. We will increase the page size as we have to store additional indices (increased storage - probably less of an issue but I thought I'd add this as a heads up)
> 2. We could have a detrimental effect on writes since each write will also have to update the indices.
> 
> Just thought I'd add these as a comment just so we could keep an eye on any write performance degradation or if we see a uptick in RDS storage usage.

@wendelfabianchinsamy We are doing more reading than writing to the table, we will monitor after deployment if there are any performance issues, but this should improve our read speed.

wendelfabianchinsamy (approved):

wendelfabianchinsamy (approved):

db/migrate/20240209113122_remove_globus_uuid_index.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveGlobusUuidIndex < ActiveRecord::Migration[6.1]
+  def change
+    remove_index :allocator, column: :globus_uuid
```
ashwinisukale, line 5: I guess we might need this index, because we are using this column in serializer at many places, you can take a look at this commit for reference https://github.com/datacite/lupo/commit/eeb811c70c10b33c634d61f9e9ed23e9b352993c

ashwinisukale (changes requested):
I have one comment regarding `globus_uuid` column index. Otherwise code looks good.

db/migrate/20240209113122_remove_globus_uuid_index.rb
kaysiz: @ashwinisukale unfortunately we are not using the column to fetch any data(that is either `.where`, `find_by` or `order_by`), inside the serialiser it is used to conditionally include the the attribute.

ashwinisukale (approved):

PR #1126, opened 2024-02-14
Add/Remove Indexes - Media Table

kaysiz:
## Purpose

We have an incomplete index that is causing the query to do a filesort in the filter. The first part of the query uses index to fetch record but other part is now using filesort.
```sql
=> EXPLAIN for: SELECT `media`.* FROM `media` WHERE `media`.`dataset` = 38 ORDER BY created DESC
+----+-------------+-------+------------+------+-----------------------------------+-----------------+---------+-------+------+----------+---------------------------------------+
| id | select_type | table | partitions | type | possible_keys                     | key             | key_len | ref   | rows | filtered | Extra                                 |
+----+-------------+-------+------------+------+-----------------------------------+-----------------+---------+-------+------+----------+---------------------------------------+
|  1 | SIMPLE      | media | NULL       | ref  | dataset_updated,FK62F6FE44D3D6B1B | dataset_updated | 8       | const |    1 |    100.0 | Using index condition; Using filesort |
+----+-------------+-------+------------+------+-----------------------------------+-----------------+---------+-------+------+----------+---------------------------------------+
```

closes: #1124

## Approach

- Add a composite index that will cover both parts of the query
```sql
=> EXPLAIN for: SELECT `media`.* FROM `media` WHERE `media`.`dataset` = 38 ORDER BY created DESC
+----+-------------+-------+------------+------+---------------------------------------------------------------+-----------------------------+---------+-------+------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys                                                 | key                         | key_len | ref   | rows | filtered | Extra       |
+----+-------------+-------+------------+------+---------------------------------------------------------------+-----------------------------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | media | NULL       | ref  | dataset_updated,FK62F6FE44D3D6B1B,index_media_dataset_created | index_media_dataset_created | 8       | const |    1 |    100.0 | Using where |
+----+-------------+-------+------------+------+---------------------------------------------------------------+-----------------------------+---------+-------+------+----------+-------------+
```
- Refactor `Media` related queries

## Types of changes


app/models/media.rb
```diff
@@ -50,7 +50,7 @@ def doi_id=(value)
 
   def set_defaults
     current_media =
-      Media.where(dataset: dataset).order("media.created DESC").first
+      Media.where(dataset: dataset).order("created DESC").first
```
kaysiz, line 53: Specifying a table name is only necessary on complex queries, simple queries like this one you can just specify the column as there is no ambiguity as to where the column belongs

app/models/doi.rb
```diff
@@ -1874,7 +1874,7 @@ def metadata_version
   end
 
   def current_media
-    media.order("media.created DESC").first
```
kaysiz, line 1877: This is already defined in the association `has_many :media, -> { order "created DESC" }..` as a result it produces a query like this `"SELECT `media`.* FROM `media` WHERE `media`.`dataset` = 38 ORDER BY created DESC, media.created DESC"`

wendelfabianchinsamy (approved):

ashwinisukale (approved):

PR #1202, opened 2024-06-29
Return dois from other registration agencies through REST API (optionally)

jrhoads:
## Purpose
There are many dois in the database and elasticsearch that are not originally registered by datacite.
These are unavailable through the current REST API.  They are available through the GraphQL api.
We would like to make these available as read-only through the REST api.

## Approach
Add an optional parameter that enables the inclusion of dois from other registration agencies.

## Types of changes


richardhallett (approved):
Based on conversations internally, this is a reasonable compromise to allow searching using REST API over non DataCite DOIs without having to create more complicated new endpoints.
Long term this probably needs to be reworked to consider how search happens for our API endpoints when also mixed in with registration services.

PR #1205, opened 2024-07-02
Use SQS_PREFIX not ES_PREFIX for queues

richardhallett:
## Purpose
Now that I want to remove the use of ES_PREFIX as we have a specific new ES server domain.
Then some detection needs to change, namely as ES_PREFIX will now be empty we need a way to choose queues.

## Approach

The way im working this is to add a new SQS_QUEUE which I will then set configured in terraform.
You will note that I'm still use a weird detection for some salesforce message part, this is because this entire code is very strange and I'm not sure I want to break it further, so for now I just use SQS_QUEUE as the way to identify here.

I do note that while Rails has technically two separate environments setup for stage and test, we're not actually using this properly, I'm afraid of making that change as it seems more involved, but is probably ideally the better solution for identifying the environment.

N/A

## Learning
N/A

## Types of changes

### DIFF
--- a/app/models/concerns/indexable.rb
+++ b/app/models/concerns/indexable.rb
@@ -32,11 +32,11 @@ module Indexable
       elsif instance_of?(Event)
         OtherDoiJob.perform_later(dois_to_import)
       # ignore if record was created via Salesforce API
-      elsif instance_of?(Provider) && !from_salesforce && (Rails.env.production? || ENV["ES_PREFIX"] == "stage")
+      elsif instance_of?(Provider) && !from_salesforce && (Rails.env.production? || ENV["SQS_PREFIX"] == "stage")
         send_provider_export_message(to_jsonapi.merge(slack_output: true))
-      elsif instance_of?(Client) && !from_salesforce && (Rails.env.production? || ENV["ES_PREFIX"] == "stage")
+      elsif instance_of?(Client) && !from_salesforce && (Rails.env.production? || ENV["SQS_PREFIX"] == "stage")
         send_client_export_message(to_jsonapi.merge(slack_output: true))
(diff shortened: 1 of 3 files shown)


app/models/concerns/indexable.rb
```diff
@@ -32,11 +32,11 @@ module Indexable
       elsif instance_of?(Event)
         OtherDoiJob.perform_later(dois_to_import)
       # ignore if record was created via Salesforce API
-      elsif instance_of?(Provider) && !from_salesforce && (Rails.env.production? || ENV["ES_PREFIX"] == "stage")
+      elsif instance_of?(Provider) && !from_salesforce && (Rails.env.production? || ENV["SQS_PREFIX"] == "stage")
```
richardhallett, line 35: See the PR Description, but yes this isn't much better, but arguably not worse

digitaldogsbody (approved):
One question: is the rails env definitely called `test` (as opposed to `testing`)? I don't remember, hence the question, but if it is `testing` then I think some of this will fall over (such as the SQS queue names).

Other than that, happy for merge :)

richardhallett:
It's `test` but it's not set to that, the test environments are set to `stage` which is not helpful. Hence why there was some code in there that was doing t hings like if Rails.env.stage? but then also asking ES_PREFIX because it can't differentiate between staging and test.

So for the container config for stage SQS_PREFIX will be "stage"
And for test SQS_PREFIX will be "test"
And for production it will just fall back to environment and be "production"

I did think maybe I could just not fall back to the Rails.env entirely and just be explicit about it, but I sort of went for a half and half compromise for how it was already working.

digitaldogsbody (approved):

PR #1220, opened 2024-07-25
Rework gbif cleanup

wendelfabianchinsamy:
## Purpose

## Types of changes

### DIFF
--- a/app/jobs/delete_gbif_events_job.rb
+++ b/app/jobs/delete_gbif_events_job.rb
@@ -13,14 +13,13 @@ class DeleteGbifEventsJob < ApplicationJob
     end
 
     # delete event records from mysql
-    result = Events.where(id: ids).delete_all
-    Rails.logger.info("#{label}: #{result} event records deleted")
+    sql = ActiveRecord::Base.sanitize_sql_array(["DELETE FROM events WHERE id IN (?)", ids])
+    ActiveRecord::Base.connection.execute(sql)
 
     # delete event documents from elasticsearch
     bulk_payload = ids.map { |id| { delete: { _index: index, _id: id } } }
-    response = Event.__elasticsearch__.client.bulk(body: bulk_payload)
-    Rails.logger.info("#{label}: #{response}")
+    Event.__elasticsearch__.client.bulk(body: bulk_payload)
   rescue => err
-    Rails.logger.error("#{label}: #{are.message}")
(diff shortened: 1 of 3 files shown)


app/jobs/delete_gbif_events_job.rb
```diff
@@ -3,11 +3,18 @@
...
-  def perform(id, options = {})
-    event = Event.find_by(uuid: id)
+  def perform(ids, options = {})
+    label = options[:label]
+    index = options[:index]
 
-    event.destroy! if event.present?
+    # delete event records from mysql
+    sql = ActiveRecord::Base.sanitize_sql_array(["DELETE FROM events WHERE id IN (?)", ids])
+    ActiveRecord::Base.connection.execute(sql)
```
ashwinisukale, line 12: When we delete the events using raw query, we are skipping the [`after_commit on: [:destroy]` ](https://github.com/datacite/lupo/blob/master/app/models/concerns/indexable.rb#L53)callback.

wendelfabianchinsamy: @ashwinisukale yes this is by design as I will be using the bulk api to delete the documents in the events index.

app/models/event.rb
```diff
@@ -880,6 +880,41 @@ def self.loop_through_events(options)
...
+    filter = options[:filter] || {}
+    label = options[:label] || ""
+    job_name = options[:job_name] || ""
+    query = options[:query].presence
+    delete_count = 0
+    max_delete_count = options[:max_delete_count]
+
+    response = Event.query(query, filter.merge(page: { size: 1, cursor: [] }))
+
+    if response.size.positive?
```
ashwinisukale, line 895: we should verify `if response.present? && response.respond_to?(:size) && response.size.positive?` response present, incase of an error

app/models/event.rb
```diff
@@ -880,6 +880,41 @@ def self.loop_through_events(options)
...
+      while response.size.positive? && delete_count < max_delete_count
+        response = Event.query(query, filter.merge(page: { size: size, cursor: cursor }))
+
+        break unless response.size.positive?
```
ashwinisukale, line 899: we can make little readable by checking it like this `break if response.size.zero?`

ashwinisukale (approved):
looks good, there are some refactoring changes needed, but as you said we will be deleting this code so no need to worry on refactoring part.

Thanks,

PR #1221, opened 2024-07-26
Allow multiple values in the resource-type-id parameter

digitaldogsbody:
## Purpose

We allow querying against multiple values for `resourceType`, but not `resourceTypeGeneral`. Since `resourceTypeGeneral` is a controlled list, it seems useful to allow multiple queries in a single API call, as this uses the much more efficient `filter` approach in ES (vs doing `?query=types.resourceTypeGeneral:x OR types.resourceTypeGeneral:y`)

This PR adds support for multiple values in the `resource_type_id` parameter, which ultimately uses the value from `resourceTypeGeneral` via https://github.com/datacite/lupo/blob/master/app/models/doi.rb#L1232-L1237 and https://github.com/datacite/lupo/blob/master/app/models/doi.rb#L2489-L2495

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -962,7 +962,7 @@ class Doi < ApplicationRecord
     minimum_should_match = 0
 
     filter << { terms: { doi: options[:ids].map(&:upcase) } } if options[:ids].present?
-    filter << { term: { resource_type_id: options[:resource_type_id].underscore.dasherize } } if options[:resource_type_id].present?
+    filter << { terms: { resource_type_id: options[:resource_type_id].split(",").map(&:underscore).map(&:dasherize } } if options[:resource_type_id].present?
     filter << { terms: { "types.resourceType": options[:resource_type].split(",") } } if options[:resource_type].present?
     if options[:provider_id].present?
       options[:provider_id].split(",").each { |id|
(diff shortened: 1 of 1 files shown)


jrhoads (approved):

app/models/doi.rb
```diff
@@ -962,7 +962,7 @@ def self.query(query, options = {})
     minimum_should_match = 0
 
     filter << { terms: { doi: options[:ids].map(&:upcase) } } if options[:ids].present?
-    filter << { term: { resource_type_id: options[:resource_type_id].underscore.dasherize } } if options[:resource_type_id].present?
+    filter << { terms: { resource_type_id: options[:resource_type_id].split(",").map(&:underscore).map(&:dasherize) } } if options[:resource_type_id].present?
```
ashwinisukale, line 965: Code changes look perfect, can we format it like this,

```
  if options[:resource_type_id].present?
    resource_type_ids = options[:resource_type_id]
                          .split(",")
                          .map { |id| id.strip.underscore.dasherize }
   filter << { terms: { resource_type_id: resource_type_ids } }
  end
  ```

digitaldogsbody: I've done this, thanks for the suggestion :)

ashwinisukale (approved):
Looks great, thanks for addressing the changes :)

PR #1233, opened 2024-08-29
Fix for duplicate ids appearing in connections when events contain uppercase and lowercase source_ids and target_ids

codycooperross:
## Purpose

Addresses issue where uppercase and lowercase ids in an Event's `source_id` or `target_id` would duplicate connections in a Doi record, like citations and thus `citation_ids`, `citation_count`. See the `citations` attribute here: https://api.datacite.org/dois/10.17632/579pxjyjz8.1 

```
"citations": {
"data": [
{
"id": "10.32725/jab.2020.004",
"type": "dois"
},
{
"id": "10.32725/jab.2020.004",
"type": "dois"
}
]
},
```

closes: datacite/datacite#2195 

## Approach

Adds `.compact.map(&:downcase)` before `.uniq` when calculating connections. Slightly refactors `citations_over_time` method.

Would require a re-index to be reflected on every record in the REST API. 

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -1296,7 +1296,7 @@ class Doi < ApplicationRecord
   # remove duplicate citing source dois,
   # then show distribution by year
   def citations_over_time
-    citation_events.map { |event| [event.occurred_at, event.source_doi.downcase] }.sort_by { |v| v[0] }.uniq { |v| v[1] }.
+    citation_events.pluck(:occurred_at, :source_doi).map { |v| [v[0], v[1].downcase] }.sort_by { |v| v[0] }.uniq { |v| v[1] }.
       group_by { |v| v[0].utc.iso8601[0..3] }.
       map { |k, v| { "year" => k, "total" => v.length } }.
       sort_by { |h| h["year"] }
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -1273,11 +1273,11 @@ class Doi < ApplicationRecord
   end
 
   def reference_ids
-    reference_events.pluck(:target_doi).compact.uniq.map(&:downcase)
+    reference_events.pluck(:target_doi).compact.map(&:downcase).uniq
   end
 
   def reference_count
-    reference_events.pluck(:target_doi).uniq.length
+    reference_events.pluck(:target_doi).compact.map(&:downcase).uniq.length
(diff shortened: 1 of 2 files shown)


app/models/doi.rb
```diff
@@ -1273,30 +1273,30 @@ def downloads_over_time
   end
 
   def reference_ids
-    reference_events.pluck(:target_doi).compact.uniq.map(&:downcase)
+    reference_events.pluck(:target_doi).compact.map(&:downcase).uniq
   end
 
   def reference_count
-    reference_events.pluck(:target_doi).uniq.length
+    reference_events.pluck(:target_doi).compact.map(&:downcase).uniq.length
```
wendelfabianchinsamy, line 1280: @codycooperross the only suggestion I would make is not exactly to do with your change but with the existing code repetition. This happens in a few places but if you look at the method reference_count it's code is exactly the same as reference_ids except for the additional length method invocation. Because we didn't reuse the code we had to make this change in two places. 

This is certainly outside the scope of your  PR. I will create an issue to clean up this code to make more DRY (don't repeat yourself).

I will approve this PR.

wendelfabianchinsamy (approved):

app/models/doi.rb
codycooperross: Thank you, Wendel!

app/models/doi.rb
```diff
@@ -1273,30 +1273,30 @@ def downloads_over_time
...
   def citation_count
-    citation_events.pluck(:source_doi).uniq.length
+    citation_events.pluck(:source_doi).compact.map(&:downcase).uniq.length
   end
 
   # remove duplicate citing source dois,
   # then show distribution by year
   def citations_over_time
-    citation_events.pluck(:occurred_at, :source_doi).uniq { |v| v[1] }.
+    citation_events.pluck(:occurred_at, :source_doi).map { |v| [v[0], v[1].downcase] }.sort_by { |v| v[0] }.uniq { |v| v[1] }.
```
codycooperross, line 1299: @wendelfabianchinsamy I just added back the `pluck` because I think the `map` would have selected the entire Event record and may not have been performant for DOIs with many events. All good?

wendelfabianchinsamy (approved):

PR #1252, opened 2024-09-25
Add new UAT environment

richardhallett:
For usage to explicitly define the test environment that is for members

## Purpose
This will let us set the test environment that is used by users to be something more explicit and seperate from the stage environment.
Practical benefit is we can then simplify code in some palces and use Rails.env as intended.

## Approach
Add new env config

N/A

## Learning
N/A

## Types of changes

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -26,3 +26,6 @@ production:
 
 stage:
   <<: *defaults
+
+uat:
+  <<: *defaults
--- a/config/environments/uat.rb
+++ b/config/environments/uat.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+Rails.application.configure do
+  # Settings specified here will take precedence over those in config/application.rb.
+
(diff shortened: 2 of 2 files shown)


kaysiz:
@richardhallett why do we need this environment? Seems like already have `test` and `staging`

richardhallett:
@kaysiz 
To clarify further "test" is not test in terms of DataCite Test Environment, test is for local testing that is used by rspec
Therefore in fact the current DataCite Test Environment is actually set the Rails environment to staging for both Our development staging environment (i.e. api.stage.datacite.org) and the user testing environment (api.test.datacite.org).

So the main reason is to simplify code and be explicit when referring to the correct environment in the rails application.

kaysiz (approved):

wendelfabianchinsamy (approved):

jrhoads (approved):

PR #1272, opened 2024-11-08
Chang default DOI search order

jrhoads:
## Purpose
Users reported that the search results were coming back in an order they were not expecting.  It turns out that the default search order had been set to the date of last update in the database.  This was discussed and determined that the relevancy of the search results or _score in ES/OS should have been the default (with other sort options available if specified in the API call).

closes: datacite/datacite/issues/2206

## Approach
Sets the default sort to be based on _score of search results, descending.

## Types of changes

### DIFF
--- a/app/controllers/datacite_dois_controller.rb
+++ b/app/controllers/datacite_dois_controller.rb
@@ -50,7 +50,7 @@ class DataciteDoisController < ApplicationController
       when "relevance"
         { "_score": { "order": "desc" } }
       else
-        { updated: { order: "desc" } }
+        { "_score": { "order": "desc" } }
       end
 
     page = page_from_params(params)
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -2518,7 +2518,7 @@ class Doi < ApplicationRecord
 
     def update_publisher_from_string
       self.publisher_obj = { name: publisher_before_type_cast }
-     end
+    end
 
     def reset_publishers
       self.publisher_obj = nil
--- a/app/models/doi/graphql_query.rb
+++ b/app/models/doi/graphql_query.rb
@@ -4,7 +4,7 @@ module Doi::GraphqlQuery
   class Builder
     include Modelable
 
-    DEFAULT_CURSOR = [0, ""]
+    DEFAULT_CURSOR = ["Infinity", 0]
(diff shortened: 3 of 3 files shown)


jrhoads:
@codycooperross, @MaryHirsch and @KellyStathis

Are there any documentation or support implications of this change.
It makes the default sorting for the REST and GraphQL APIs based on the relevancy of the search.
I would argue that this is what is expected.

I could not find documentation on the default search sorting order when I looked.  I eventually discovered in code that it had been set to the date of last update in REST (most recent first) and date of creation (oldest first) in GraphQL

codycooperross:
@MaryHirsch @KellyStathis Has the sort order of the REST API come up in community consultations or support requests?

MaryHirsch:
I can't think of any examples of this coming up in support.

KellyStathis:
@codycooperross I agree that for current REST API users, it may be unexpected. However, I couldn't find anywhere where we documented a default sort for the REST API. I think the benefits of sorting by relevance outweigh this.

Another place to check is the DOI list sorting in Fabrica. We document that the default is Date Updated: e.g. https://support.datacite.org/docs/fabrica-member-dashboard#dois. If that's specified in the query with `sort=-updated` (vs. the omission of the sort parameter) then this should be unaffected; otherwise, we would need to change that to explicitly sort by date.

jrhoads:
I ran a quick test on the staging server for Fabrica with developer tools open to the network tab.  It looks like that by default when you got to the /dois page, it will send out a rest query with `sort` param set empty.  The interface has the option to select a sort order. But by default it will send an empty sort.  If the user sends a query at all in the search bar, then the `sort` parameter gets set to relevance.

digitaldogsbody:
It makes sense to me that we preserve the advertised Fabrica behaviour (both because it's what people are used to/is documented, but also because if there's no search, sorting by recency seems like the most useful behaviour) so we should adopt Kelly's suggestion and make the matching change in Fabrica to enforce recency sort when there is no query.

digitaldogsbody (approved):
Happy for this to be merged once Bracco has been updated to explicitly request recency sort when loading the list of DOIs for the first time

wendelfabianchinsamy (approved):

jrhoads:
Updated Fabrica.  Will merge this later today

PR #1299, opened 2024-12-06
Refactors `/dois` faceting and adds support for on-demand facets

codycooperross:
## Purpose

Refactors approach to OpenSearch aggregations and serialized faceting at `/dois` to reduce calls to OpenSearch for unused facets. Implements new `facets=` URL param that allows the requestor to select from the possible facets. 

closes: https://github.com/datacite/product-backlog/issues/127 https://github.com/datacite/lupo/issues/1301

## Approach

Builds the aggregations based on the requested facets, default or otherwise, before assembling the OpenSearch query. Accepts comma-separated values via a new `facets=` URL param that selects from the possible facets. Facets are serialized based on supported and available aggregations in the OpenSearch response. 

## Types of changes

### DIFF
--- a/app/controllers/datacite_dois_controller.rb
+++ b/app/controllers/datacite_dois_controller.rb
@@ -136,6 +136,7 @@ class DataciteDoisController < ApplicationController
           source: params[:source],
           scroll_id: params[:scroll_id],
           disable_facets: disable_facets,
+          facets: params[:facets],
           page: page,
           sort: sort,
           random: params[:random],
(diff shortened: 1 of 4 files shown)


bklaing2 (approved):

app/models/doi.rb
```diff
@@ -674,82 +674,137 @@ def as_indexed_json(_options = {})
...
-        published: {
-          date_histogram: {
-            field: "publication_year",
-            interval: "year",
-            format: "year",
-            order: {
-              _key: "desc",
-            },
-            min_doc_count: 1,
+  def self.all_doi_aggregations
```
jrhoads, line 677: At this point it could be a constant.
```
AGGREGATION_DEFINITIONS = {...
...
}
```

app/models/doi.rb
```diff
@@ -674,82 +674,137 @@ def as_indexed_json(_options = {})
...
+      :views,
+      :downloads,
+      :citations
+    ]
+  end
+
+  def self.query_aggregations(disable_facets: false, facets: nil)
+    return {} if disable_facets
+
+    requested_facets = facets.respond_to?(:split) ? facets.split(",").map(&:strip).map(&:underscore).map(&:to_sym).uniq : default_doi_query_facets
```
jrhoads, line 806: This assumes that facets is only allowed to be a string.  It could be useful to allow arrays of symbols to be passed in.  And treat the string as a special case.

app/models/doi.rb
```diff
@@ -674,82 +674,137 @@ def as_indexed_json(_options = {})
...
+    requested_facets.index_with { |facet| all_doi_aggregations.dig(facet) }.compact
```
jrhoads, line 807: I think this is similar to the slice method, available on hashes.
so this would become
`all_doi_aggregations.slice(*requested_facets)`
or
`AGGREGATION_DEFINITIONS .slice(requested_facets)`

keys that are not in the original hash are discarded.

app/controllers/datacite_dois_controller.rb
jrhoads, line 1: This is great work.  Some time in the future we may want to refactor this out in to some sort of DoiAggregationSerializer.

app/models/doi.rb
codycooperross: @jrhoads Do you have thoughts on where the constant should be defined?

jrhoads: It can be a constant within the model.

PR #1326, opened 2025-03-03
New reference repositories endpoint

bklaing2:
## Purpose
The existing `/clients` endpoint doesn't return the complete data that is returned from a repository in GraphQL. In order to be able to migrate away from GraphQL, this PR exposes the `ReferenceRepository` model used for GraphQL in a new endpoint `/reference-repositories`

## Approach
Created a new controller `ReferenceRepositoriesController` with `index` and `show` methods, as well as a `ReferenceRepositorySerializer`. These expose the existing `ReferenceRepository` model (used in the graphQL) in the REST API.

- [ ] I added aggregations to the `ReferenceRepository.find_by_id()` method in order for the controller's `index` method to function when passing the `id/ids` param. This may not be ideal since the `show` method also uses `find_by_id` but doesn't need aggregations.
  - For reference, the `ClientController` uses `Client.where()` when setting `@client` for the `show` method.
  - If we simply remove the support for the `id(s)` param we can remove the aggregations from `find_by_id`, but that would break consistency with other endpoints.

## Types of changes


spec/requests/reference_repositories_spec_GQL.rb
jrhoads, line 1: Is this needed?

bklaing2: Oops no it's not

jrhoads (approved):

PR #1337, opened 2025-05-06
Move web service to profiles for explicit startup

richardhallett:
This is a different way of specifying what comes up with docker compose

This is a breaking workflow change as by default with "docker compose up" it will not start the rails app "web"

## Purpose

This is to allow for example local running of the rails app but still starting anxillery services needed by the app i.e. mysql, opensearch etc.
This is the same as using the docker compose file .local

If this PR is accepted seperately we might wish to remove the .local one.

## Approach
Uses docker profiles https://docs.docker.com/compose/how-tos/profiles/

N/A

## Learning
See approach

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,13 @@ Using Docker.
 docker run -p 8065:80 datacite/lupo
 ```
 
-or
+or using with docker compose `app` profile
 
 ```bash
-docker-compose up
+docker compose --profile app up
 ```
+
+Without specifying the `app` profile just the core infrastructure parts will start i.e. mysql, opensearch etc
 
 If you want to build the docker image locally (instead of pulling it from docker hub)
  and use docker compose for development you can use
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,6 @@
 services:
   web:
+    profiles: [ app ]
     env_file: .env
     environment:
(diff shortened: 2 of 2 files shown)


digitaldogsbody:
Because I'm an awkward sod, I'm curious if this can be inverted? I.e the default is to start all services and the profile flag is used to prevent the rails container. 

This feels like a more intuitive approach to me, in terms of keeping what is (I assume) the more common usage as the "base" operation. 

I haven't looked at the docs for how the compose profiles work, so this idea may not actually be practical.

digitaldogsbody:
I just know I'm going to forget about this by the time I next spin up the lupo stuck locally and get very confused 😅

richardhallett:
I appreciate the concern, unfortunatly it's not too easy to do the opposite, because you can't specify profiles to disable.
There is a possible trick that you can specify the default profile using an empty string like "" and if you did it like that, then I could say the default is everything and still specify a profiles for just say "services".
This however might not work on all docker compose versions, it's not documented officially as far as I can tell, just some people reporting this as a possibility.

So I think maybe I won't merge this PR. I'll think about it some more and see if there is an alternative compromise.

The main goal of this was an example to show not requiring special docker compose files, but that perhaps is just the easiest solution.

PR #1339, opened 2025-05-16
Move metadata file storage for metadata history to S3

richardhallett:
## Purpose

This PR is to move the historical metadata storage from the metadata table in the database to an object store, namely S3.

The reasoning behind this is for performance and cost optimisations see: https://github.com/datacite/datacite/issues/2329 and sub issues

## Approach

Using the AWS SDK store the XML in S3 instead of storing directly in the xml blob in RDS. The database instead will just have the object_key for lookup.
We handle this by changing the model to fetch this data from S3 if it's got an object_key stored.
Most likely in the future once we migrate all historical data we can simplify this to always pull from S3.

See also: https://github.com/datacite/docs-engineering/blob/main/Planning/metadata_storage/Metadata%20storage%20-%20Stage%201%20Phase%201%20-%20Notes.md

- Is this still performant enough? This isn't necessarily pre merge but will need further testing most likely on staging.

## Learning

- Testing of how to handle S3 object keys, researching ideas on how best to store, what libraries to consider, considerations of long term maintainability.

## Types of changes

### DIFF
--- a/app/models/metadata.rb
+++ b/app/models/metadata.rb
@@ -1,23 +1,12 @@
 # frozen_string_literal: true
 
+require "aws-sdk-s3"
+
 class Metadata < ApplicationRecord
   include Bolognese::Utils
   include Bolognese::DoiUtils
 
   include Cacheable
-
-  # added a getter and setter for the xml attribute utf-8 encoding was not working as expected and we had failing specs.
-  # the failing spec spec/models/metadata_spec.rb:48
-  # example 1: Céline was persisted as C\xC3\xA9line.
-  # example 2: PatiÃ±o was persisted as Pati\xC3\x83\xC2\xB1o
-  def xml=(value)
-    value = value&.force_encoding("UTF-8")
-    super(value)
-  end
-
-  def xml
-    super&.force_encoding("UTF-8")
-  end
 
   alias_attribute :created_at, :created
   alias_attribute :datacite_doi_id, :doi_id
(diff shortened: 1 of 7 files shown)


app/models/metadata.rb
```diff
@@ -29,15 +18,53 @@ def xml
...
+    # example 2: PatiÃ±o was persisted as Pati\xC3\x83\xC2\xB1o
+    # See also the forced return on the getter
+    value = value&.force_encoding("UTF-8")
+    super(value)
+  end
+
+  def xml
+    # We check if the value is already set to object_key i.e. externally stored
+    # This helps avoid unnecessary calls to S3
+    if super == object_key && !object_key.nil?
```
wendelfabianchinsamy, line 42: to reduce code complexity and to make the code read a little more intentional could we change this line to:
`if super == object_key && !object_key.nil?`
to
`if super == object_key && object_key.present?`

richardhallett: This isn't going to reduce code complexity, it'd be a logic change, because what I'm checking for is nil, which is what I want to avoid, the comparison is then going ahead, I actually dont care if it's empty string "" because the other check would fail anyway.

Now granted you might say well why not do a "emptiness" check to avoid trying to checking against empty string, but I'm not sure it's giving a specific advantage.

app/models/metadata.rb
```diff
@@ -29,15 +18,53 @@ def xml
...
   end
 
   def doi_id
-    doi.doi
+    # Don't try and access nil doi object
+    # Can occur for example when building the metadata object and the relation
+    # hasn't yet been setup
+    unless doi.nil?
+      doi.doi
+    end
```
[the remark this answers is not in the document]
richardhallett: I don't particularly like the safe navigation operator but i appreciate it's not unusual Ruby syntax, so sure I can change it.

app/models/metadata.rb
wendelfabianchinsamy: @richardhallett Thank you. Yes I agree nil does seem to be more expressive with regards to the intent. So I'll resolve the conversation. What I will say is in turns on Rails, object_key.present? is the more idiomatic way of performing this check. But I appreciate that this is just semantics and either way does not diminish the code. It would, however, be nice for us to strive for idiomatic solutions and thus ensuring that we solve solutions the "rails way".

wendelfabianchinsamy (approved):

PR #1341, opened 2025-05-21
Implements JSON Schema for titles

codycooperross:
## Purpose

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -106,16 +106,22 @@ class Doi < ApplicationRecord
   validates_presence_of :doi
   validates_presence_of :url, if: Proc.new { |doi| doi.is_registered_or_findable? }
 
-  json_schema_validation = {
-    message: ->(errors) { errors },
-    schema: PUBLISHER_JSON_SCHEMA
-  }
-
   def validate_publisher_obj?(doi)
     doi.validatable? && doi.publisher_obj? && !(doi.publisher_obj.blank? || doi.publisher_obj.all?(nil))
   end
 
-  validates :publisher_obj, if: ->(doi) { validate_publisher_obj?(doi) }, json: json_schema_validation
+  def validate_field?(field_name)
+    validatable? && self[field_name].present?
+  end
+
+  def self.json_schema_validation(schema_name)
+    {
+      message: ->(errors) { errors },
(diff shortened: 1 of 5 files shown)


richardhallett:
What's this actually trying to solve? I'm not sure I follow the original ticket.

Validation should ideally happen in one place, which currently is bolognese, if this doesn't work properly then the fix is there. Ideally currently titles are already validated against the XSD to validate the title data.

I'd worry also that we're going to get into a mess of how we handle different schema versions in the database, this schema validation here is going to be for latest schema.

The data stored in the database is really just text, the fact it's json is largely irrelevant.

Also my thoughts on restructuring the database leans to removing us storing individual metadata data in separate database columns entirely, this needs fleshing out, but currently it does limit us to how the data is structured.

Happy to talk about this more as perhaps this is a bit too broad to discuss on the PR.

app/models/doi.rb
svogt0511, line 1: Minor comment: PUBLISHER_JSON_SCHEMA goes away in favor of schema_file_path (https://github.com/datacite/lupo/blob/27518e14ffd00ae8ce7e9edc7d0a4af51fe433f0/app/models/doi.rb#L8)

app/models/doi.rb
```diff
@@ -106,16 +106,19 @@ class Doi < ApplicationRecord
...
-    message: ->(errors) { errors },
-    schema: PUBLISHER_JSON_SCHEMA
-  }
-
   def validate_publisher_obj?(doi)
     doi.validatable? && doi.publisher_obj? && !(doi.publisher_obj.blank? || doi.publisher_obj.all?(nil))
   end
 
-  validates :publisher_obj, if: ->(doi) { validate_publisher_obj?(doi) }, json: json_schema_validation
+  def validate_json_attribute?(attribute)
```
svogt0511, line 113: Is it possible to modify publisher validation to use validate_json_attribute?

app/models/schemas/doi/titles.json
svogt0511, line 1: We require at least one element in the titles array. Does this schema specify that requirement?  Would an empty array be ok?

svogt0511 (commented):
Looks good, I just have a few  minor comments.

PR #1346, opened 2025-06-05
Remove AWS from tests

jrhoads:
## Purpose
Removes direct AWS client calls from testing and replaces them with a service object, allowing for better test isolation without requiring AWS credentials.

## Approach
This change introduces a new service object, `SqsMessageService`, to handle interactions with AWS SQS. This centralizes the SQS logic and allows for stubbing the AWS client during testing. Additionally, an initializer is added to configure the AWS SDK for testing without needing real credentials.

### Key Modifications
- Removed AWS access key, secret access key, and region from the `parallel_ci.yml` workflow.
- Created a new service object `SqsMessageService` to encapsulate SQS message sending logic.
- Modified the `Indexable` concern to use the new `SqsMessageService`.
- Added a new initializer `aws_sdk_test_setup.rb` to configure the AWS SDK for testing with stubbed responses and dummy credentials.
- Updated `rails_helper.rb` to use `Aws.config[:region]` for constructing the SQS host in VCR configurations when in the test environment.

- The `SqsMessageService` uses `Aws.config[:stub_responses]` to determine whether to construct a dummy SQS URL or use the result from `sqs.get_queue_url` when running tests.
- The `aws_sdk_test_setup.rb` initializer is specifically configured for the test environment and sets `stub_responses: true` and uses dummy credentials.

## Types of changes

### DIFF
--- a/app/services/sqs_message_service.rb
+++ b/app/services/sqs_message_service.rb
@@ -1,0 +1,35 @@
+class SqsMessageService
+  def self.call(body, options = {})
+    sqs = Aws::SQS::Client.new # Assumes Aws.config is set globally
+    
+    queue_name_prefix = ENV["SQS_PREFIX"].present? ? ENV["SQS_PREFIX"] : Rails.env
+    target_queue_name = "#{queue_name_prefix}_#{options[:queue_name]}"
+    
+    effective_queue_url = if Aws.config[:stub_responses]
+                            # If stubbing is active, always construct a well-formed dummy URL.
+                            # The actual result of sqs.get_queue_url is unreliable when stubbed (returns "String").
+                            dummy_url = "https://sqs.#{Aws.config[:region]}.amazonaws.com/000000000000/#{target_queue_name}"
+                            # Call get_queue_url for its potential side effects or future compatibility, but ignore its direct return for the URL.
+                            sqs.get_queue_url(queue_name: target_queue_name) 
+                            dummy_url
+                          else
+                            # For real calls, use the result from get_queue_url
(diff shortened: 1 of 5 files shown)


app/services/sqs_message_service.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class SqsMessageService
```
wendelfabianchinsamy, line 3: I love this. Abstracting away implementation details will make our code easier to maintain since these types of things will only need to change in one place. 

Couple of thoughts:
1. Could you rename the service object SqsMessenger? This is a more idiomatic rails naming convention.
2. Could we rename the call method to enqueue? I feel like this may help with code readability. Since the intention is more easily understood without having to look at the call method to understand it's purpose.

@jrhoads Any thoughts? I am very much open to different opinions and interpretations.

jrhoads: > I love this. Abstracting away implementation details will make our code easier to maintain since these types of things will only need to change in one place.
> 
> Couple of thoughts:
> 
> 1. Could you rename the service object SqsMessenger? This is a more idiomatic rails naming convention.
I've seen it both ways.  I think that since Services are a new concept in this codebase, it is useful to have it in the name of the object.

> 2. Could we rename the call method to enqueue? I feel like this may help with code readability. Since the intention is more easily understood without having to look at the call method to understand it's purpose.
I agree.  That will make it much clearer what it is actually doing.

> @jrhoads Any thoughts? I am very much open to different opinions and interpretations.

wendelfabianchinsamy (approved):

PR #1349, opened 2025-06-13
Set default test environment variables

jrhoads:
## Purpose
This change aims to improve the reliability and consistency of the test environment by setting reasonable default values for required environment variables. This prevents test failures due to missing configuration when environment variables are not explicitly set.

## Approach
The approach taken is to explicitly define default values for a comprehensive set of environment variables within the `spec/rails_helper.rb` file. This ensures that even without external configuration, the test suite has the necessary information to run successfully. Dummy JWT keys are also generated if not already present to support testing of JWT-based authentication.

### Key Modifications

- Added default values for numerous environment variables in `spec/rails_helper.rb`, including database connection, API endpoints, service credentials, and various application settings.
- Included logic to generate dummy JWT private and public keys for testing purposes if they are not already set as environment variables.

- The default environment variables are set using `ENV["VARIABLE_NAME"] ||= "default_value"`, which only assigns the default if the variable is not already defined.
- The JWT key generation uses the `openssl` library to create a 2048-bit RSA key pair for testing.

## Types of changes

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -4,9 +4,72 @@ ENV["RAILS_ENV"] = "test"
 ENV["TEST_CLUSTER_NODES"] = "1"
 ENV["PREFIX_POOL_SIZE"] = "20"
 
+# Set reasonable default environment variables for testing if not already set
+ENV["TEST_ENV_NUMBER"] ||= "" # For parallel tests, often set by CI, default to empty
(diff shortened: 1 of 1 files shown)


richardhallett (commented):
In general fine. Did you consider putting the env variables in the test.rb environment specific file instead of rails helper?
Not sure I've really thought which is better, but given it's specific config for environment then maybe makes sense to override there.

richardhallett (approved):
No strong opinion

PR #1395, opened 2025-09-19
Fix: Sync indices during reindex with `index_sync_enabled?`

jrhoads:
## Purpose
This PR addresses an issue where indices can become out of sync during large-scale reindexing operations by introducing a mechanism to control index synchronization. The core of this change revolves around the `index_sync_enabled?` flag.

## Approach
The change introduces a new service, `AppSettings`, to manage the `index_sync_enabled?` flag, which is stored in the Rails cache. The `IndexJobDoiRegistration` job now checks this flag before performing a secondary indexing operation. If the flag is enabled, the DOI is indexed into its inactive index as well, ensuring that both the active and inactive indices are synchronized. A new rake task is also added to easily enable, disable, and check the status of index syncing.

### Key Modifications

*   **`IndexJobDoiRegistration`:** Modified to include a check for `index_sync_enabled?` on the object being indexed. If enabled, it performs an additional `index_document` call targeting the object's inactive index.
*   **`Doi` Model:** Added `index_sync_enabled?` instance and class methods for easier access to the synchronization status.
*   **`AppSettings` Service:** New service created to manage the `INDEX_SYNC_ENABLED` flag in the Rails cache, providing methods to enable, disable, and check its status.
*   **`lib/tasks/index_sync.rake`:** New rake task file added to provide commands for managing the index sync status (`enable`, `disable`, `status`).
*   **`spec/services/app_settings_spec.rb`:** Added comprehensive tests for the `AppSettings` service to ensure correct behavior of the index sync flag management.

*   The `index_sync_enabled?` check in `IndexJobDoiRegistration` is bypassed in the test environment to prevent unintended behavior during tests.
*   The `AppSettings` service uses `Rails.cache` to persist the `INDEX_SYNC_ENABLED` flag.
*   The `AppSettings.index_sync_enabled?` method performs a strict boolean comparison (`== true`) to ensure that only explicitly `true` values enable synchronization. Any other value (including `nil` or string "true") will result in `false`.

## Types of changes

### DIFF
--- a/app/jobs/index_job_doi_registration.rb
+++ b/app/jobs/index_job_doi_registration.rb
@@ -10,8 +10,19 @@ class IndexJobDoiRegistration < ApplicationJob
     Rails.logger.error error.message
   end
 
+  def index_sync_enabled?(obj)
+    if Rails.env.test?
+      return false
+    end
+    obj.respond_to?(:index_sync_enabled?) && obj.index_sync_enabled?
+  end
+
   def perform(obj)
     response = obj.__elasticsearch__.index_document
     Rails.logger.error "[Elasticsearch] Error #{response.inspect}" unless %w(created updated).include?(response["result"])
+    if index_sync_enabled?(obj)
+      response2 = obj.__elasticsearch__.index_document(index: obj.inactive_index)
+      Rails.logger.error "[Elasticsearch] Error #{response2.inspect}" unless %w(created updated).include?(response2["result"])
+    end
   end
 end
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -2491,6 +2491,14 @@ class Doi < ApplicationRecord
     read_attribute("publisher_obj")
   end
 
+  def index_sync_enabled?
+    self.__class__.index_sync_enabled?
+  end
+
+  def self.index_sync_enabled?
+    Rails.cache.read("INDEX_SYNC_ENABLED") == true
+  end
+
   def self.repair_landing_page(id: nil)
     if id.blank?
       Rails.logger.error "[Error] No id provided."
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -2496,7 +2496,7 @@ class Doi < ApplicationRecord
   end
 
   def self.index_sync_enabled?
-    Rails.cache.read("INDEX_SYNC_ENABLED") == true
+    AppSettings.index_sync_enabled?
   end
 
   def self.repair_landing_page(id: nil)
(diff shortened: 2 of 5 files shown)


richardhallett (commented):
The idea in general is if we've set a flag all indexing jobs that might have been queued are going to do an additional index into the inactive
So 
1. This is going to have an overhead every time we have to index while this is active, remembering and given this is for reindexing, it will mean that during this period overall we'll have a higher write count on OS, especially depending on load from users. Is this acceptable? Have we considered the impact?
2. What about the other times we index, this is only capturing a subset of indexing right?

I'm not convinced that having an inactive/active is a good idea in general still, it's useful when we might want to change the structure, but if we're just reindexing from source data, then perhaps just doing it to the live index is acceptable, this avoids this problem. Needs more thought to consider trade offs.

digitaldogsbody:
> I'm not convinced that having an inactive/active is a good idea in general still, it's useful when we might want to change the structure, but if we're just reindexing from source data, then perhaps just doing it to the live index is acceptable, this avoids this problem. Needs more thought to consider trade offs.

My concern would be that doing such a lot of writes to the active index is going to severely impact the speed of queries against that index, and could potentially slow the production API to uselessness whilst the reindex is happening. We might be able to mitigate it with some of the OS settings to control how often field-indices are updated etc but we definitely should do some investigation and testing.

jrhoads:
> Another thing I'd be curious to know is what happens in the case of conflicting index writes and/or race conditions?
> 
> Let's say that the "reindex everything" job reads the record from the database and then the record is updated and a "reindex one" job is called whilst the "reindex everything" is retrieving other records to populate its bulk request to OS. Could the "reindex one" job run at the same time, or before, the bulk request containing the older version of the record?
> 
> If so, this could lead to a situation where the OS data ends up being incorrect in the inactive index, and it might be worth us thinking about how we are going to handle that.

@digitaldogsbody I've modified the behavior so that instead of initiating a write at the same time as the active index, we enqueue the job for the inactive index.  This should reduce the likelihood of race conditions.

digitaldogsbody (approved):
Looks good, thanks for addressing the race condition feedback - even though I agree that it's unlikely, as you mentioned, I'm nonetheless happier that it should now be handled gracefully.

I think in the future it's worth revisiting our general approach to inactive indexes, along the lines that Richard was talking about. I wonder if we could do away with them _except_ when we need to make structure changes, and then when that's required, use the inbuilt OS functionality to copy directly from one index to another.

Until then, this is a good solution to the issue, so happy for the merge.

PR #1397, opened 2025-09-26
Add rake task to delete dois for a client

kaysiz:
## Purpose

Remove DiSSCo Doi records from the test system and reduce storage ceiling

closes: https://github.com/datacite/product-backlog/issues/423

## Approach

Create a rake task that takes client_id(symbol), start and end date, and delete dois for that client within the date range 

- [ ] Notify DiSSCo that the records will be removed and suggest a timeframe. 
- [ ] In the agreed timeframe, remove the records from the database and OpenSearch. 
- [ ] Reduce the storage ceiling for OpenSearch in test.

## Types of changes

### DIFF
--- a/lib/tasks/client.rake
+++ b/lib/tasks/client.rake
@@ -209,4 +209,45 @@ namespace :client do
       puts "Client prefix for client #{target.symbol} and prefix #{prefix} created."
     end
   end
+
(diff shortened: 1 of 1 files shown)


codycooperross:
Given that this is an operation that we won't do in production, would it make sense to do an environment check before proceeding?

PR #1421, opened 2025-10-30
Enforce data structure and data type consistency for JSON metadata

svogt0511:
## Purpose

closes: https://github.com/datacite/product-backlog/issues/325

## Approach

See https://github.com/datacite/lupo/pull/1341 for the approach

Some notes:

- A number of DOI fields are expected to be arrays of objects, but users will sometimes submit a request with an object instead of an array of one object.  Lupo tries to correct this using Array.wrap, but not in every case.  Consequently users can submit incorrect json data, but a DOI is still created because lupo corrects it.  The json validations check for this situation and will report errors against this type of data if lupo has not already corrected the incoming data.  Some of these fields are: ...  Other's that flag the issue via json schema are: creators

## Types of changes

## Summary by CodeRabbit

* **New Features**
  * Expanded JSON Schema coverage for DOI metadata (updated draft) with richer controlled vocabularies and new schema validations.

* **Refactor**
  * Validation now resolves schemas dynamically and centralizes JSON-schema checks for many metadata fields.

* **Chore**
  * Upgraded a JSON validation dependency to a newer version series.

* **Tests**
  * Fixtures and specs adjusted to array-based metadata shapes and stricter validation/error expectations.


jrhoads (approved):

jrhoads (approved):
Other than the comment, all good and even that is not a functional change, just a cleanup.

app/models/schemas/doi/related_identifier.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "else": {
+    "not": {
+      "anyOf": [
+        { "required": ["relatedMetadataScheme"] },
+        { "required": ["schemeUri"] },
+        { "required": ["schemeType"] }
+      ]
+    }
+  },
```
KellyStathis: @svogt0511 This restriction _is_ in the schema documentation—that 12.c, 12.d, and 12.e should only be used when 12.b = "HasMetadata" or "IsMetadataFor": https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/relatedidentifier/

However, I don't think it is in the XSD as XSD doesn't support conditional requirements like this.

In the interest of maintaining parity between the XSD and the JSON Schema, could we remove this?

app/models/schemas/doi/name_identifier.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "nameIdentifier": {
+      "type": ["string", "null"]
+    },
+    "nameIdentifierScheme": {
+      "type": ["string", "null"]
+    }
+  },
+  "dependentRequired": {
+    "affiliationIdentifier": ["nameIdentifierScheme"]
```
KellyStathis: @svogt0511 Should this line be as follows?

`"nameIdentifier": ["nameIdentifierScheme"]`

As opposed to `"affiliationIdentifier": ["nameIdentifierScheme"]`?

KellyStathis (commented):
Hi @svogt0511, I noted a couple things from reading through this but haven't tested yet. Happy to take a closer look when I'm back next week!

app/models/schemas/doi/affiliation.json
```diff
@@ -0,0 +1,23 @@
...
+      "type": ["string", "null"]
+    },
+    "name": {
+      "type": ["string", "null"]
+    },
+    "schemeUri": {
+      "type": ["string", "null"]
+    }
+  },
+  "dependentRequired": {
```
codycooperross: While logical, this is inconsistent with the XSD definition, which does not suggest dependent values between affiliationIdentifier and affiliationIdentifierScheme:

```
<xs:complexType name="affiliation">
<xs:annotation>
<xs:documentation>Uniquely identifies an affiliation, according to various identifier schemes.</xs:documentation>
</xs:annotation>
<xs:simpleContent>
<xs:extension base="nonemptycontentStringType">
<xs:attribute name="affiliationIdentifier" type="xs:string" use="optional"/>
<xs:attribute name="affiliationIdentifierScheme" type="xs:string" use="optional"/>
<xs:attribute name="schemeURI" type="xs:anyURI" use="optional"/>
</xs:extension>
</xs:simpleContent>
</xs:complexType>

File: app/models/schemas/doi/name_identifier.json (hunk 2 above)
Comment (codycooperross) [reply app/models/schemas/doi/name_identifier.json]: nameIdentifier also cannot be null since it has `nonemptycontentStringType`.

File: app/models/schemas/doi/publication_year.json [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+      "maxLength": 4
+    },
+    {
+      "type": "integer",
+      "minimum": 1000,
+      "maximum": 3000,
+      "inclusiveMinimum": true,
+      "inclusiveMaximum": true
+    }
+  ],
```
codycooperross: Should publicationYear be required by this schema?

app/models/doi.rb
```diff
@@ -146,6 +151,34 @@ def validate_publisher_obj?(doi)
...
+  validates :subjects, if: proc { |doi| doi.validate_json_attribute?(:subjects) }, json: { message: ->(errors) { errors }, schema: lambda { schema_file_path("subjects") } }, unless: :only_validate
+  validates :contributors, if: proc { |doi| doi.validate_json_attribute?(:contributors) }, json: { message: ->(errors) { errors }, schema: lambda { schema_file_path("contributors") } }, unless: :only_validate
+  validates :dates, if: proc { |doi| doi.validate_json_attribute?(:dates) }, json: { message: ->(errors) { errors }, schema: lambda { schema_file_path("dates") } }, unless: :only_validate
+  validates :raw_language, if: proc { |doi| doi.validate_json_attribute?(:raw_language) }, json: { message: ->(errors) { errors }, schema: lambda { schema_file_path("language") } }, unless: :only_validate
+  validates :resource_type, if: proc { |doi| doi.validate_json_attribute?(:resource_type) }, json: { message: ->(errors) { errors }, schema: lambda { schema_file_path("resource_type") } }, unless: :only_validate
```
codycooperross: Should this validate the `types` property? That's where resourceType and resourceTypeGeneral are stored.

app/models/schemas/doi/date.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "dateInformation": {
+      "type": "string"
+    }
+  },
+  "required": [
+    "date"
+  ],
+  "dependentRequired": {
+    "date": ["dateType"]
```
codycooperross: Should `dateType` just be required? It seems like it is that way in the XSD:

```
<xs:element name="dates" minOccurs="0">
<xs:complexType>
<xs:sequence>
<xs:element name="date" minOccurs="0" maxOccurs="unbounded">
<xs:annotation>
<xs:documentation>Different dates relevant to the work.</xs:documentation>
<xs:documentation>YYYY,YYYY-MM-DD, YYYY-MM-DDThh:mm:ssTZD or any other format or level of granularity described in W3CDTF. Use RKMS-ISO8601 standard for depicting date ranges.</xs:documentation>
</xs:annotation>
<xs:complexType>
<xs:simpleContent>
<xs:extension base="xs:string">
<xs:attribute name="dateType" type="dateType" use="required"/>
<xs:attribute name="dateInformation" use="optional"/>
</xs:extension>
</xs:simpleContent>
</xs:complexType>
</xs:element>
</xs:sequence>
</xs:complexType>
</xs:element>
```

app/models/schemas/doi/alternate_identifier.json
```diff
@@ -0,0 +1,17 @@
...
+    "alternateIdentifier": {
+      "type": "string"
+    },
+    "alternateIdentifierType": {
+      "type": "string"
+    }
+  },
+  "additionalProperties": false,
+  "dependentRequired": {
+    "alternateIdentifier": ["alternateIdentifierType"]
```
codycooperross: Should both `alternateIdentifier` and `alternateIdentifierType` be required?

app/models/schemas/doi/related_identifier.json
codycooperross: Agreeing with Kelly—let's remove this conditional since it's not imposed by the XSD.

app/models/schemas/doi/related_identifier.json
```diff
@@ -0,0 +1,58 @@
...
+  "dependentRequired": {
+    "relatedIdentifier": ["relatedIdentifierType", "relationType"]
```
codycooperross: I'll stop mentioning this going forward because there's a clear pattern and likely a good reason behind it, but just wondering if it's worth having these as simply `required` rather than `dependentRequired`

app/models/schemas/doi/formats.json
```diff
@@ -0,0 +1,8 @@
+{
+  "title": "Formats",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "array",
+  "items": {
+    "$ref": "size.json"
```
codycooperross: I don't think it makes a difference since sizes and formats are both just arrays of strings, but this points to size.json rather than format.json

app/models/schemas/doi/funding_reference.json
```diff
@@ -0,0 +1,35 @@
+{
+  "title": "FundingReference",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "funderName": {
```
codycooperross: It looks like funderName cannot be null or empty:

```
<xs:element name="funderName" minOccurs="1" maxOccurs="1">
<xs:annotation>
<xs:documentation>Name of the funding provider.</xs:documentation>
</xs:annotation>
<xs:simpleType>
<xs:restriction base="nonemptycontentStringType"/>
</xs:simpleType>
</xs:element>
```

Should this be `"type": ["string"]` ?

app/models/schemas/doi/related_items.json
```diff
@@ -0,0 +1,9 @@
+{
+  "title": "RelatedItems",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "array",
+  "minItems": 0,
```
codycooperross: Also a clear pattern with `"minItems": 0`, but are these necessary?

codycooperross (changes requested):
Took a pass on the JSON Schemas aside from relatedItem—these look great! I made some comments where there might be inconsistencies with the XSD or the current JSON representation.

Will get to relatedItem soon!

app/models/schemas/doi/contributor.json
```diff
@@ -0,0 +1,42 @@
...
+      "oneOf": [
+        {
+            "$ref": "controlled_vocabularies/contributor_type.json"
+        },
+        {
+          "type": "null"
+        }
+      ]
+    },
+    "lang": {
```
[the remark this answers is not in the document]
KellyStathis: That's interesting, it should be allowed. In the XML, it is attached to contributorName: https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/#id2

There is `<xs:attribute ref="xml:lang"/>` in the XSD.

app/models/schemas/doi/affiliation.json
KellyStathis: Yes, I agree with this. Although logic is described in the Schema documentation at https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/creator/#b-affiliationidentifierscheme - "If affiliationIdentifier is used, affiliationIdentifierScheme is mandatory." - it isn't part of the XSD, so we should omit it from the JSON Schema validation.

app/models/schemas/doi/related_identifier.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "then": {
+    "properties": {
+      "relatedMetadataScheme": {
+        "type": "string"
+      },
+      "schemeUri": {
+        "type": "string"
+      },
+      "schemeType": {
```
KellyStathis: I believe `schemeType` can be null. It is in this example: https://support.datacite.org/docs/api-create-dois#request-payload-1

app/models/schemas/doi/contributor.json
```diff
@@ -0,0 +1,42 @@
+{
+  "title": "Contributor",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "name": {
```
KellyStathis: This is an interesting one. I was testing with metadata pulled from some recently created/updated DOIs, to see if this validation would have impacted them.

I found this DOI which in the JSON has a contributor `givenName` and `familyName`, but not `name`: https://api.datacite.org/dois/10.34804/supra.2021092825

```json
"contributors": [
{
"nameType": "Personal",
"givenName": "Jacopo",
"familyName": "Torrisi",
"affiliation": [],
"contributorType": "DataManager",
"nameIdentifiers": [
{
"nameIdentifier": "",
"nameIdentifierScheme": "ORCID"
}
]
}
],
```

Using this metadata to create a DOI on staging failed with this error:
```json
{
    "errors": [
        {
            "source": "contributors",
            "title": "Object at `/0` is missing required properties: name",
            "uid": "10.1111/742r-wc63"
        }
    ]
}
```

From my understanding of the XSD, `contributorName` is required even if `givenName` and `familyName` are provided. And the corresponding XML for this DOI does have a `contributorName`:

```xml
<contributors>
    <contributor contributorType="DataManager">
      <contributorName nameType="Personal">Torrisi, Jacopo</contributorName>
      <givenName>Jacopo</givenName>
      <familyName>Torrisi</familyName>
      <nameIdentifier nameIdentifierScheme="ORCID" schemeURI=""/>
      <affiliation affiliationIdentifierScheme="ROR"/>
    </contributor>
  </contributors>
```

How is that contributorName being generated for the XML? I am just thinking through the potential impact on user who are currently providing not providing `contributor.name`, but are providing `contributor.givenName` and `contributor.familyName`, if we introduce this.

KellyStathis: Tagging @codycooperross for input into this as well :)

app/models/schemas/doi/related_identifier.json
```diff
@@ -0,0 +1,58 @@
...
+    "relatedIdentifierType": {
+      "$ref": "controlled_vocabularies/related_identifier_type.json"
+    },
+    "relatedMetaDataScheme": {
+      "type": "string"
+    },
+    "schemeUri": {
+      "type": "string"
+    },
+    "resourceTypeGeneral": {
```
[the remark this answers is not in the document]
codycooperross: Since the API currently accepts null values for many fields in my testing, yes, the JSON Schema should permit null values for all fields that are not `xs:string` or `nonemptycontentStringType` in the XSD.

app/models/schemas/doi/related_identifier.json
```diff
@@ -0,0 +1,64 @@
...
+        "type": "string"
+      },
+      "schemeType": {
+        "type": "string"
+      },
+      "relationTypeInformation": {
+        "type": "string"
+      }      
+    }
+  },
```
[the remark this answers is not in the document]
KellyStathis: > The then schema only defines the types of relatedMetadataScheme, schemeUri, schemeType, and relationTypeInformation; it never requires them. When relationType is HasMetadata or IsMetadataFor, these fields remain optional instead of mandatory.

We should **not** do this, these fields aren't conditionally required in the XSD and we don't want to make the JSON validation stricter than the XSD.

app/models/schemas/doi/alternate_identifier.json
svogt0511: Yes.  They should both be required.  

Question: The metadata.xsd shows:

Affiliation: <xs:extension base="nonemptycontentStringType">
-versus-
AlternateIdentifier: <xs:extension base="xs:string">

Shouldn't they both be nonemptycontentStringType?

app/models/schemas/doi/contributor.json
svogt0511: This was always very confusing to me. 

As I understand it, contributorName is always required.

If nameType = Unknown then Contributor can have a given name, family name and a completely independent name.

If nameType = Organization then Contributor has any name you want to give it. (**Although, if you are converting the type to organization to one of the other types, givenName and familyName are preserved in the DB (and probably in elasticSearch.  A topic for another time.)

If nameType = Person, the Name field is generated from the givenName and familyName.

I would not remove the name requirement.  I would not back out our requirements to fit incorrect data.

I thought that if there are incorrect dois in the DB, that the user would be prodded to correct them if/when the doi is updated, otherwise, there is not much we could do, except perhaps, to flag them as being incorrect at some point.

Does this make sense?

svogt0511: I also think I really have to look over the json schema again for this one.  It is a bit complicated.

app/models/schemas/doi/alternate_identifier.json
KellyStathis: @codycooperross `alternateIdentifier` is an optional property. Can you clarify your suggestion that it should be required here?

Yes, there are definitely inconsistencies like this. So technically, an alternateIdentifier can have length 0 characters.

app/models/schemas/doi/date.json
svogt0511: I added dateType to the required fields.

app/models/schemas/doi/geo_location.json
```diff
@@ -0,0 +1,24 @@
...
+      "$ref": "geo_location_box.json"
+    },
+    "geoLocationPlace": {
+      "$ref": "geo_location_place.json"
+    },
+    "geoLocationPolygon": {
+      "$ref": "geo_location_polygon.json"
+    }
+  },
+  "required": [
```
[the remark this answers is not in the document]
codycooperross: Still not sure if these are necessary here.

codycooperross (approved):

PR #1514, opened 2026-04-15
Remove random password generator

### LINKED ISSUE #1512: Remove random password suggestion functioanlity
We have a /random endpoint that is using a very old gem to provide random strings that are only used by Bracco for the "Suggest Password" feature.

Since we are removing that (Issue: https://github.com/datacite/product-backlog/issues/774 PR: https://github.com/datacite/bracco/pull/960) and @kaysiz already indicated that there were issues with upgrading the underlying gem to work with the latest Ruby and Rails versions, it seems sensible to remove this endpoint and associated code rather than reimplementing it.

digitaldogsbody:
## Purpose
The random password functionality is being removed from its only consumer (bracco) so we can get rid of it here to.
Bracco issue: https://github.com/datacite/product-backlog/issues/774
Bracco PR: https://github.com/datacite/bracco/pull/960
Lupo issue: https://github.com/datacite/lupo/issues/1512

I've also done the same work for @kaysiz's Rails 8 upgrade branch: https://github.com/datacite/lupo/pull/1513

closes: https://github.com/datacite/lupo/issues/1512

**Important**: the bracco PR needs to be released to production first to ensure that this functionality is not still being used

## Types of changes


digitaldogsbody:
I will fix the issue with the Gemfile.lock as soon as I have an internet connection stable enough to build the container without failing :grimacing:

Gemfile.lock
```diff
@@ -916,4 +910,4 @@ DEPENDENCIES
   webmock (~> 3.18, >= 3.18.1)
 
 BUNDLED WITH
-   2.6.9
+   2.3.20
```
jrhoads: Will this cause an issue?

digitaldogsbody: Good catch, thanks - this is because I used `bundle lock` locally rather than building the container and then copying the lockfile and I forgot to change to the correct Ruby environment. It shouldn't cause an issue but I'll update bundler and regenerate the lockfile to be safe.

jrhoads (approved):

digitaldogsbody:
The bracco PR is merged and in production so this is ok to release

PR #1537, opened 2026-06-05
Update software enum in openapi.yaml

KellyStathis:
## Purpose

## Approach

Adds latest RSP additions to "software" enum, to sync with list in Fabrica.

## Types of changes


codycooperross (approved):

KellyStathis:
@codycooperross I've just caught this up to the latest changes from main and added another recent RSP addition to the list. Could you please re-review and merge for the next lupo release? Thanks!

codycooperross:
@KellyStathis Since we’re planning for this to become a free text field, I’m wondering if we should continue to update this as if it’s a controlled list. What do you think?

KellyStathis:
@codycooperross I think it still makes sense for Fabrica: https://github.com/datacite/bracco/pull/974 - we're still planning to have the suggested options for the [Integrations from Registered Service Providers](https://support.datacite.org/edit/service-provider-software-integrations), as part of the program.

I agree about openapi.yaml, but my preference would be to keep them in sync for now until we make the free text change.

codycooperross:
Agreed on Fabrica, and sounds good. Noting that we'll need to update this as a developer reference if/when these enums can no longer be expected.

codycooperross (approved):

PR #1541, opened 2026-06-16
Update bolognese

bklaing2:
## Purpose
Update the version of bolognese to the latest. The latest version of bolognese removes many normalizations. Some tests needed to be updated to reflect these changes.

Two bolognese functions used in lupo were renamed, which have been updated where called.

closes: https://github.com/datacite/product-backlog/issues/465

## Types of changes


Gemfile.lock
```diff
@@ -864,4 +864,4 @@ DEPENDENCIES
   webmock (~> 3.26, >= 3.26.2)
 
 BUNDLED WITH
-   2.6.9
```
codycooperross: Not sure if it's a concern but because it's been mentioned to me in the past for other repos: you may want to use the version of bundler already in the Gemfile.lock. You can use a specific version of bundler like this: `bundler _2.6.9_ update`

app/models/doi.rb
```diff
@@ -2599,9 +2592,9 @@ def update_language
   def update_field_of_science
```
codycooperross: @bklaing2 I'm not sure if this method actually does anything anymore, and the callback that calls it might be unnecessary now. Can you double-check that? Previously, it added a FOS subjects, but I believe that functionality is now removed from bolognese, and therefore these name_to_subject and hsh_to_subject methods don't need to be called.

bklaing2: @codycooperross I checked it out and I think it is still necessary. Even though the FOS normalization doesn't occur anymore, the [bolognese methods](https://github.com/datacite/bolognese/blob/e9e0bee9305089c68bd7448234459ea7ce805649/lib/bolognese/utils.rb#L1334) still sanitize and format the inputs.

codycooperross: Sounds good—thanks for checking!

codycooperross (approved):
Looks good to me, but might be good to have someone else on the dev team review just in case.

spec/requests/datacite_dois/post_spec.rb
```diff
@@ -590,7 +587,7 @@
                                                                         "edition" => "1",
                                                                         "contributors" => [
                                                                             "contributorType" => "ProjectLeader",
-                                                                            "name" => "Hallett, Richard",
+                                                                            "name" => "Richard, Hallett",
```
jrhoads, line 590: Is this right?

spec/concerns/crosscitable_spec.rb
```diff
@@ -835,7 +786,7 @@
       expect(meta["creators"].first).to eq(
         "familyName" => "Fenner",
         "givenName" => "Martin",
-        "name" => "Fenner, Martin",
+        "name" => "Martin Fenner",
```
jrhoads, line 789: Some appear to come back  as FAMILY, GIVEN
others GIVEN, FAMILY
And this one is GIVEN FAMILY

Was there a normalization step before that is no longer applied?
Is it just keeping what they send in for "name"

bklaing2: @jrhoads we have removed normalizations from bolognese and the changes to the tests reflect that

spec/requests/datacite_dois/post_spec.rb
jrhoads: Resolved from other resposne

jrhoads (approved):

spec/concerns/crosscitable_spec.rb
bklaing2: In this case specifically it is now strictly based on the input data, and no longer tries to extract and format the names

spec/requests/datacite_dois/post_spec.rb
codycooperross: Yes, based on the XML: 
```
      <contributors>
        <contributor contributorType="ProjectLeader">
            <contributorName nameType="Personal">Richard, Hallett</contributorName>
            <givenName>Richard</givenName>
            <familyName>Hallett</familyName>
        </contributor>
      </contributors>
```

PR #1551, opened 2026-06-24
Match more permissive detail= conditional to serializer

codycooperross:
## Purpose

The DataciteDoiSerializer uses `if: Proc.new { |_object, params| params && params[:detail] }`, so the same conditional is now used in the controller so they're equally permissive. (OAI-PMH, i.e. viringo, uses `detail=True`, which calls for this change.) 

## Types of changes

### DIFF
--- a/app/controllers/datacite_dois_controller.rb
+++ b/app/controllers/datacite_dois_controller.rb
@@ -181,7 +181,7 @@ class DataciteDoisController < ApplicationController
       end
 
       # Pluck XML from DB if params[:detail] is true
-      if params[:detail] == "true"
+      if params[:detail]
         doi_ids = results.map { |r| r._source["doi"] }.compact
         xml_by_doi = Doi.where(doi: doi_ids).pluck(:doi, :xml).to_h
 


digitaldogsbody (approved):
Whilst I'm happy to approve this as a temporary fix - I would rather that in the long run we transitioned to proper support for the parameter's value both here and in the serializer, because having identical behaviour for `?detail=true` and `?detail=false` is a bit daft in my opinion.

Other possibilities for right now could be to fix Viringo to use `true` instead of `True`, or switching this statement to allow for both values, either by direct comparison or (better) doing a lower case comparison.

codycooperross:
> Whilst I'm happy to approve this as a temporary fix - I would rather that in the long run we transitioned to proper support for the parameter's value both here and in the serializer, because having identical behaviour for `?detail=true` and `?detail=false` is a bit daft in my opinion.
> 
> Other possibilities for right now could be to fix Viringo to use `true` instead of `True`, or switching this statement to allow for both values, either by direct comparison or (better) doing a lower case comparison.

There are a few places where we check for param truthiness instead of a specific value, and it’s bit uneven throughout lupo. Marked as a separate piece of work, which would include an analysis of usage across our repos as well: 

- https://github.com/datacite/product-backlog/issues/873

PR #1553, opened 2026-06-24
Removes XML from OS documents and implements DB XML retrieval for GraphQL 

codycooperross:
## Purpose

This PR follows: 

- https://github.com/datacite/lupo/pull/1540

... as part of a general strategy for removing XML from OpenSearch documents for performance and storage considerations. 

The PR hydrates XML in GraphQL requests if XML is requested. The check is relatively simple and checks the GQL query string for `xml`, but this should weed out most requests that do not request XML. 

With GQL XML hydration in place, XML is no longer necessary in DOI OS documents, so this PR also removes the `xml` attribute from the Doi model's `as_indexed_json`.

The following steps would be re-indexes to remove XML from the DOIs indices, noted here: https://github.com/datacite/product-backlog/issues/867

closes: https://github.com/datacite/product-backlog/issues/865
closes: https://github.com/datacite/product-backlog/issues/866

## Types of changes

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -686,7 +686,6 @@ class Doi < ApplicationRecord
       "fields_of_science" => fields_of_science,
       "fields_of_science_repository" => fields_of_science_repository,
       "fields_of_science_combined" => fields_of_science_combined,
-      "xml" => xml,
       "is_active" => is_active,
       "landing_page" => landing_page,
       "agency" => agency,
--- a/spec/requests/datacite_dois/datacite_dois_spec.rb
+++ b/spec/requests/datacite_dois/datacite_dois_spec.rb
@@ -249,6 +249,7 @@ describe DataciteDoisController, type: :request, vcr: true do
       expect(json["data"].size).to eq(10)
       json["data"].each do |doi|
         expect(doi.dig("attributes")).to include("xml")
+        expect(doi.dig("attributes", "xml")).to be_present
       end
     end
 
--- a/spec/requests/datacite_dois/datacite_dois_spec.rb
+++ b/spec/requests/datacite_dois/datacite_dois_spec.rb
@@ -70,6 +70,9 @@ describe DataciteDoisController, type: :request, vcr: true do
       expect(last_response.status).to eq(200)
       expect(json["data"].size).to eq(10)
       expect(json.dig("meta", "total")).to eq(10)
+      json["data"].each do |doi|
+        expect(doi.dig("attributes", "xml")).not_to be_present
+      end
     end
 
     it "returns dois with scroll" do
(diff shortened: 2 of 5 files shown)


kaysiz (approved):

app/models/doi.rb
```diff
@@ -686,7 +686,6 @@ def as_indexed_json(_options = {})
       "fields_of_science" => fields_of_science,
       "fields_of_science_repository" => fields_of_science_repository,
       "fields_of_science_combined" => fields_of_science_combined,
-      "xml" => xml,
```
jrhoads, line 689: This removes it from the JSON document sent to OpenSearch.
Do we also want to remove it from the mappings?

jrhoads: Not a blocker for this PR

spec/requests/datacite_dois/datacite_dois_spec.rb
```diff
@@ -249,6 +249,7 @@ def clear_doi_index
       expect(json["data"].size).to eq(10)
       json["data"].each do |doi|
         expect(doi.dig("attributes")).to include("xml")
+        expect(doi.dig("attributes", "xml")).to be_present
```
jrhoads, line 255: Do we test the opposite?
That the XML attribute is not present if detail does not equal true?

jrhoads (approved):

spec/requests/datacite_dois/datacite_dois_spec.rb
digitaldogsbody: No, because the current behaviour of the serializer is just check the _existence_ of the query parameter `detail` and not its value, so `?detail=false` actually produces the same behaviour.

See https://github.com/datacite/lupo/pull/1551

app/models/doi.rb
digitaldogsbody: I would say not right now, because this allows us to test at production scale in a slightly safer way.

Leaving the mapping the same right now means not needing to drop the index-template that is used in OS for both the active and inactive indices and means we can avoid a full reindex until this approach is better validated at the much larger scale and traffic of the production system.

If it turns out not to work and we have to quickly switch back to retrieving the XML from OS, then we can revert the changes but we will only have to reindex the fraction of records that have changed between deployment of this and the revert, rather than all of them, so we'll be back up and running much faster.

Once we're in a position where we are happy that the XML-from-db is working consistently well in production, then we can drop the field from the index, reindex everything in the inactive index and switch.

spec/requests/datacite_dois/datacite_dois_spec.rb
codycooperross: The serialization logic related to the detail param was largely unchanged, so I hadn't added a test for this originally. Added now in spec/requests/datacite_dois/datacite_dois_spec.rb

app/models/doi.rb
codycooperross: Added this as a to-do here:

- https://github.com/datacite/product-backlog/issues/877

jrhoads (approved):

PR #1556, opened 2026-07-07
Embed classic MDS protocol into Lupo

kaysiz:
## Summary

- Embeds the classic DataCite MDS protocol (formerly Poodle) in-process on dedicated hosts (`mds.datacite.org` / test/stage/local), so MDS URLs keep working while Lupo remains the system of record.
- Host-constrained routes serve plain-text `/doi`, `/metadata`, `/media`, and `/heartbeat` when `MDS_ENABLED` is on and the request host is in `MDS_HOSTS`; production stays off unless explicitly enabled; test/dev use `mds.local` only so REST specs are not captured.
- Thin protocol controllers over existing Lupo domain (`DataciteDoi`, CanCan, ParamsSanitizer, Bolognese format detection, shared `RequestCredentials`, domain landing-URL helpers). Single DOI upsert path for PUT `/doi` and PUT `/metadata`; no parallel Operations/Result stack.

## Configuration

- `MDS_ENABLED` (default false)
- `MDS_HOSTS` (comma-separated; falls back to production-like defaults when blank and enabled)
- `MDS_URL`, `MDS_REALM`
- Nginx template gains a dual `server_name` for MDS hosts

## Test plan

- [ ] `bundle exec rspec spec/lib/mds_spec.rb spec/routing/mds_routing_spec.rb spec/requests/mds/`
- [ ] Confirm non-MDS host still serves REST `/dois` and does not return MDS plain-text bodies for `/doi`
- [ ] Smoke: Basic auth success/failure on `mds.local` for GET/PUT `/doi`, PUT/GET/DELETE `/metadata`, POST/GET `/media`
- [ ] Smoke: draft delete allowed; findable delete returns 405
- [ ] Verify production image does not enable MDS unless `MDS_ENABLED=true` is set
- [ ] Optional: REST `GET .../dois/:id/get-url` still resolves landing URLs (shared domain policy)


app/controllers/mds/metadata_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+          params[:doi_id],
+          data: data,
+          from: from,
+          number: params[:number],
+        )
+      fail Mds::Error.new("DOI not found", status: 404) if doi_id.blank?
+
+      xml_b64 = data.present? ? Base64.strict_encode64(data) : nil
+      attrs =
+        ParamsSanitizer.new(
```
jrhoads, line 46: Could we verify that the DOI in the submitted metadata matches `params[:doi_id]` before upserting?

`DoiMinter#resolve_doi_id` returns the path DOI immediately when it is valid, without checking the identifier embedded in the XML. As a result, a request such as `PUT /metadata/10.x/a` with XML identifying `10.x/b` could store metadata for one DOI under another DOI. Please add the same path/body consistency validation used by the `/doi` endpoint and cover the mismatch case with a request spec.

app/services/mds/doi_minter.rb
```diff
@@ -0,0 +1,48 @@
...
+    def resolve_doi_id(str, data:, from:, number: nil)
+      doi = validate_doi(str)
+      return doi if doi.present?
+
+      if from == "datacite"
+        doi = doi_from_xml_identifier(data)
+        return doi if doi.present?
+      end
+
+      mint_unique_doi(str, number: number)
```
jrhoads, line 21: Could we handle a missing or invalid minting prefix explicitly before calling `generate_random_dois`?

The `POST /metadata` route permits requests without a DOI path parameter, but when the body does not provide a usable DOI or prefix, `mint_unique_doi` can call `generate_random_dois` with a blank value. That eventually reaches `validate_prefix` and may raise an internal exception instead of returning a deterministic MDS client error. Please validate the required minting input and add coverage for empty and malformed mint requests.

app/controllers/mds/media_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      render_mds("#{@media.media_type}=#{@media.url}")
+    end
+
+    def create
+      authorize! :update, @doi
+
+      data = request.raw_post
+      fail Mds::Error.new("Media type and URL missing", status: 400) if data.blank?
+
+      media_type, url = data.to_s.split("=", 2)
```
jrhoads: Could we validate both parsed values before creating the `Media` record?

The body is split on the first `=`, but blank or malformed values are still accepted—for example, `application/pdf=` or `=https://example.org/file.pdf`. Please reject missing/blank media types and URLs, and validate or decode them consistently with the MDS protocol before persisting the record. It would also be useful to add request specs for these malformed bodies.

kaysiz:
### CI flake fix (shard 14)

Latest push (`a2665603`) addresses the intermittent `parallel-test (14, 16)` failure:

```
NoMethodError: undefined method 'uid' for nil
  app/models/client.rb (Client#assign_prefix)
```

**Not MDS-related** — it showed up in `repository_type_spec` / `reference_repositories_spec` setup (`create(:client)` after a fixed `10.17616` prefix).

**Root cause:** `cached_prefix_response` could return a stale Prefix primary key after destroy/recreate of the same uid (memcached in test + `before`/`after :all`). `assign_prefix` then built a `ProviderPrefix` with an orphan FK and crashed on `provider_prefix.prefix.uid`.

**Fix:**
1. `Cacheable#cached_prefix_response` — cache only the numeric id, reload and verify uid, invalidate on mismatch
2. `Client#assign_prefix` — use `create!`, prefer the known `Prefix#uid`, nil-guard instead of raising

This is shared infrastructure (same code path on master); this long-lived PR hit the flake more often via shard packing / random order.

app/controllers/mds/metadata_controller.rb
kaysiz: Addressed in `5dae62c9` — `Mds::DoiMinter#resolve_doi_id` now compares path DOI vs DataCite XML identifier (same rule/message as PUT `/doi`: `doi parameter does not match doi of resource`). Covered in `spec/requests/mds/metadata_spec.rb` (path/body mismatch).

app/services/mds/doi_minter.rb
kaysiz: Addressed in `de4fd9a6` — `mint_unique_doi` now validates the prefix with `validate_prefix` before `generate_random_dois` and raises `IdentifierError` (`No valid prefix found` → MDS 400). Request coverage for empty/malformed mint on `POST /metadata` in `spec/requests/mds/metadata_spec.rb`.

app/controllers/mds/media_controller.rb
kaysiz: Addressed in `5a3bc63c` — `parse_media_body` rejects blank media type or URL (e.g. `application/pdf=` / `=https://…`) with 400 before persist, so blank type no longer falls through to `Media#set_defaults`. Specs in `spec/requests/mds/media_spec.rb`.

config/routes.rb
```diff
@@ -1,6 +1,32 @@
...
+      post "metadata/:doi_id", to: "metadata#create", constraints: { doi_id: /.+/ }
+      put "metadata/:doi_id", to: "metadata#create", constraints: { doi_id: /.+/ }
+      get "metadata/:doi_id", to: "metadata#show", constraints: { doi_id: /.+/ }
+      get "metadata", to: "metadata#show"
+      delete "metadata/:doi_id", to: "metadata#destroy", constraints: { doi_id: /.+/ }
+
+      resources :dois, path: "/doi", only: %i[index show update destroy],
+                       constraints: { id: /.+/ } do
+        resources :media, only: %i[index show create destroy]
+      end
```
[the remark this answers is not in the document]
kaysiz: Good catch on the nested param name.

Poodle uses the same nesting (`resources :dois` + nested `resources :media`) with only the parent `:id` constrained. Top-level `/media/:doi_id` already had `doi_id: /.+/`.

Fixed in `9437d471`: nested media now has `constraints: { doi_id: /.+/ }` so slashy DOIs match (routing spec covers GET/POST `/doi/10.14454/abc/media`). Protocol paths unchanged — routing hardening only.

app/controllers/mds/metadata_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+          {
+            doi: doi_id,
+            xml: xml_b64,
+            should_validate: true,
+            source: "mds",
+            event: "show",
+            client_id: client_symbol,
```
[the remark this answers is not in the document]
kaysiz: Agreed that `show` can move `registered` → `findable` in Lupo AASM — and that is intentional **Poodle parity**, not an oversight.

Classic Poodle `Metadatable#create_metadata` sends exactly this to REST:

```ruby
"event" => "show",
"source" => "mds",
"should_validate" => "true",
```

(and delete metadata uses `"event" => "hide"`, PUT `/doi` uses `"event" => "publish"` — same as this embed).

We are moving Poodle in-process, not redesigning MDS state transitions. Changing to `register` would diverge from production MDS/Poodle behavior (including the historical “metadata resubmit can unhide” path). Left as `show` with an in-code comment pointing at Poodle in `9437d471`.

jrhoads:
This all looks good as fas as I am concerned.
I'm going to bring in @richardhallett as @kaysiz transitions out of this work.
I'm cautious to proceed until richard is comfortable.

PR #1582, opened 2026-08-17
Implement source id of enrichments data model

wendelfabianchinsamy:
## Purpose

## Types of changes


app/models/schemas/enrichment/enrichment.json
codycooperross: These definitions can probably be excluded now that `contributors` and `resources` aren't present, right?

wendelfabianchinsamy: thank you @codycooperross. i've updated the pr.

codycooperross: :D Looks good.

codycooperross: Actually, are all of the definitions here not needed? I'm pretty sure the whole attribute can be removed.

wendelfabianchinsamy: @codycooperross that is correct. i've removed the entire block.

jrhoads (approved):

PR #1584, opened 2026-08-20
Remove software enum from openapi.yaml (now allows free text)

KellyStathis:
## Purpose

The repository (client) "software" field now allows free text, implemented in https://github.com/datacite/product-backlog/issues/647. API users should no longer expect values to conform to a set list.

closes: https://github.com/datacite/product-backlog/issues/930

## Approach

As discussed in https://github.com/datacite/lupo/pull/1537, remove the software enum and allow any string.

## Types of changes

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -191,7 +191,7 @@ paths:
         - in: query
           name: software
           schema:
-            $ref: '#/components/schemas/software'
+            type: string
         - in: query
           name: client-type
           schema:
@@ -439,7 +439,6 @@ paths:
                             $ref: '#/components/schemas/repositoryType'
                         software:
                           type: string
-                          $ref: '#/components/schemas/software'
                         isActive:
                           description: Set to true if this account can register and update DOIs.
                           type: boolean
@@ -575,7 +574,6 @@ paths:
                             $ref: '#/components/schemas/repositoryType'
                         software:
                           type: string
-                          $ref: '#/components/schemas/software'
                         isActive:
                           description: Set to true if this account can register and update DOIs.
                           type: boolean
@@ -2189,7 +2187,6 @@ components:
               type: string
             software:
               type: string
-              $ref: '#/components/schemas/software'
             subjects:
               type: array
               items:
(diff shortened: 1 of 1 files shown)


KellyStathis:
@codycooperross Would it make sense to move this out of lupo, since it is more of a documentation artifact? Not sure if we can move non-ReadMe artifacts to https://github.com/datacite/support-docs, or if there is another good home for it.

codycooperross:
I think that makes sense. @richardhallett @jrhoads Do you have any concerns with us moving the `openapi.yaml` file out of the lupo repository?

richardhallett:
I guess it's okay, I think it should live with the code as this is source of truth, but I can see the argument for moving it, though where would you move it? The support docs we already upload the file to generate the docs there.

There is an argument this file should actually be automatically generated and not manually maintained, potentially also exposed directly on the api endpoint itself, but legacy reasons it's not, also because of the way it also sort of describes metadata which makes it a little more cumbersome.

jrhoads (approved):

jrhoads:
I'm fine with this PR and I'm also ok if you want to move it into the readme.
Like @richardhallett mentioned, theoretically we should be generatign this from the code so that it doesn't stray from the implementation, but practically, it hase been edited by hand.  It can live in whichever place makes sense.

KellyStathis:
Thanks! I will see if it can be hosted in the support-docs repository: https://github.com/datacite/support-docs, otherwise it may be easiest to leave in lupo.

The only thing is I can't merge changes to lupo, even after approving reviews (don't have permissions). There is also a failed test here which I assume is unrelated to this change. @codycooperross are you able to merge this?

codycooperross:
The failing check is actually part of the reason I would like for this not to be here. Because we currently manage the file by hand, and it's user-facing documentation rather than full overview of all API functionality, I don't think it should be implicated in the development procedure. I can work on the security updates needed to merge this here for now.

KellyStathis:
I tried it out, and ReadMe's bidirectional GitHub sync ignores folders that it does not expect. Therefore, we can safely store openapi.yaml here: https://github.com/datacite/support-docs/tree/v1.8/assets. I added the file and re-imported it to the ReadMe interface without issue.

As for the question of where this _should_ live and how we generate it:

The way we currently maintain openapi.yaml, I see it as a documentation artifact for users vs. a complete representation of all API functions. For example, we make decisions around excluding certain endpoints and parameters from the file that, while technically publicly available, we don't actively support for end-users.

I do see value in automatically generated API docs for us internally, so we have a complete picture of what's supported by the REST API. That could alongside a more curated/customized artifact like what we have now.

codycooperross:
The transition to support-docs sounds good to me. Happy to try it out that way and then come back to this particular file in lupo. Shall we close this PR?

KellyStathis:
Yes sounds good, since it seems to be okay in support-docs we can close this without merging.

PR #1590, opened 2026-09-14
Use handle primary URL instead of administrative proxy for get_dois

codycooperross:
## Purpose

Addresses https://github.com/datacite/datacite/issues/2556 . The ENV variable will only be added in production, which is the environment where the bug is occurring. 

## Types of changes


richardhallett (approved):
I might have added the HANDLE_PRIMARY_URL to application.rb config as empty string so we know it exists, but I guess it's not critical, just sometimes helps with knowing what config exists.

PR #1593, opened 2026-09-16
Use HANDLE_PRIMARY_URL for paginated requests in get_dois

codycooperross:
## Purpose

Address datacite/datacite#2556

## Types of changes


app/models/concerns/helpable.rb
```diff
@@ -179,8 +179,10 @@ def get_dois(options = {})
         )
       end
 
+      base_handle_url = (ENV["HANDLE_PRIMARY_URL"].presence || ENV["HANDLE_URL"])
```
richardhallett, line 182: Is presence check to make it so it'll also ignore empty string "" ? If so fine otherwise it's not needed if nil.

richardhallett (approved):

app/models/concerns/helpable.rb
codycooperross: Yes, to enable what we talked about in Slack, i.e., setting the variable to a blank string in terraform.

